PR #2, opened 2018-03-04
Test enhancement

peter279k:
# Changed log

- add more tests for other uncovered classes.


tests/InstantTest.php
```diff
@@ -373,6 +374,34 @@ public function testIsBeforeOrEqualTo(int $s1, int $n1, int $s2, int $n2, int $c
...
+     *
+     * @param int $s1  The epoch second of the 1st instant.
+     * @param int $n1  The nanosecond adjustment of the 1st instant.
+     * @param int $s2  The epoch second of the 2nd instant.
+     * @param int $n2  The nanosecond adjustment of the 2nd instant.
+     * @param int $cmp The comparison value.
+     */
+    public function testIsBetweenExclusive(int $s1, int $n1, int $s2, int $n2, int $cmp)
+    {
+        $this->assertSame($cmp <= 0, Instant::of($s1, $n1)->isBetweenExclusive(Instant::of($s2, $n2), Instant::of($s2, $n2)));
```
BenMorel: You're not testing much in these two methods, you're only testing that the value is within an interval of 0 seconds.

What I would do here is hardcode the interval start and end, and test with a range of possible values in between:

```php
public function testIsBetweenExclusive(int $seconds, int $nanos, bool $isBetween)
{
    $this->assertSame($isBetween, Instant::of($seconds, $nanos)->isBetweenExclusive(
        Instant::of(-1, -1),
        Instant::of(1, 1)
    );
}
```

And for the provider:

```php
[
    [-1, -2, false],
    [-1, -1, false],
    [-1,  0, true],
    [-1, 1, true],
    [0, -1, true],
    [0, 0, true],
    [0, 1, true],
    [1, -1, true],
    [1, 0, true],
    [1, 1, false],
    [1, 2, false],
]
```

Same for isBetweenInclusive(), with the data provider altered accordingly.

tests/WeekOfYearTest.php
```diff
@@ -0,0 +1,31 @@
...
+    public function testCheckShouldReturnDateTimeExceptionWithFieldNotInRange()
+    {
+        WeekOfYear::check(-1);
+    }
+
+    /**
+     * @expectedException        Brick\DateTime\DateTimeException
+     * @expectedExceptionMessage Year 2000 does not have 53 weeks
+     */
+    public function testCehckTimeShouldThrowDateYimeException()
```
BenMorel: There are still 2 typos: 'Cehck' instead of 'Check' and 'Yime' instead of 'Time'!

tests/InstantTest.php
```diff
@@ -493,6 +564,28 @@ public function providerCompareTo() : array
...
+     * @return array
+     */
+    public function providerExclusiveBetweenCompareTo() : array
+    {
+        return [
+            [ 1,  0,  0,  0,  1],
+            [ 1,  0,  0,  1,  1],
+        ];
+    }
+
```
BenMorel: These providers are obsolete!

tests/DayOfMonthTest.php
```diff
@@ -0,0 +1,20 @@
...
+ */
+class DayOfMonthTest extends AbstractTestCase
+{
+    public function testCheckWithNullMonthOfYear()
+    {
+        $dayOfMonth = new DayOfMonth();
+
+        $this->assertNull($dayOfMonth->check(31));
+    }
+}
```
BenMorel: Rather than testing the `DayOfMonth` field (and testing only one edge case in this class) it would be more interesting to add a test on the calling code (in `LocalDate` or `MonthDay`) that would trigger this specific case!

tests/WeekOfYearTest.php
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Unit tests for class WeekOfYear.
+ */
+class WeekOfYearTest extends AbstractTestCase
+{
+    /**
+     * @expectedException        Brick\DateTime\DateTimeException
+     * @expectedExceptionMessage Invalid week-of-year: -1 is not in the range 1 to 53.
+     */
+    public function testCheckShouldReturnDateTimeExceptionWithFieldNotInRange()
```
BenMorel: Return => Throw

tests/PatternParserBuilderTest.php
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Unit tests for class IsoParsers.
+ */
+class PatternParserBuilderTest extends AbstractTestCase
+{
+    /**
+     * @expectedException        \RuntimeException
+     * @expectedExceptionMessage Cannot call endOptional() without a call to startOptional() first.
+     */
+    public function testEndOptionalShouldReturnRuntimeException()
```
BenMorel: Return => Throw

tests/PatternParserBuilderTest.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $patternParserBuilder = new PatternParserBuilder();
+        $patternParserBuilder->endOptional();
+    }
+
+    /**
+     * @expectedException        \RuntimeException
+     * @expectedExceptionMessage Cannot call endGroup() without a call to startGroup() first.
+     */
+    public function testEndGroupShouldReturnRunTimeException()
```
BenMorel: Return => Throw

tests/YearMonthTest.php
```diff
@@ -330,16 +330,37 @@ public function testWithYear()
         $this->assertYearMonthIs(2001, 5, YearMonth::of(2000, 5)->withYear(2001));
     }
 
+    public function testWithYearIsTheSameYear()
+    {
+        $year = (int)date('Y');
+        $month = (int)date('m');
```
BenMorel: Please use fixed year/month here

tests/YearMonthTest.php
```diff
@@ -330,16 +330,37 @@ public function testWithYear()
...
+
+        $this->assertInstanceOf(YearMonth::class, YearMonth::of($year, $month)->withYear($year));
+        $this->assertYearMonthIs($year, $month, YearMonth::of($year, $month)->withYear($year));
```
BenMorel: You should instantiate `YearMonth::of(...)` only once and use it in 2 lines.

tests/YearMonthTest.php
```diff
@@ -330,16 +330,37 @@ public function testWithYear()
...
+    }
+
     public function testWithMonth()
     {
         $this->assertYearMonthIs(2000, 12, YearMonth::of(2000, 1)->withMonth(12));
     }
 
+    public function testWithMonthWithSameMonth()
+    {
+        $month = (int)date('m');
```
BenMorel: Please use a fixed month here

tests/InstantTest.php
peter279k: I will renove this data provider.

tests/YearMonthTest.php
```diff
@@ -386,6 +407,12 @@ public function testPlusYears(int $year, int $month, int $plusYears, int $expect
         $this->assertYearMonthIs($expectedYear, $expectedMonth, $yearMonth->plusYears($plusYears));
     }
 
+    public function testPlusYearsWithParameterIsZero()
```
BenMorel: `testPlusZeroYears()`?

tests/YearWeekTest.php
```diff
@@ -395,4 +397,30 @@ public function testToString(int $year, int $week, string $expected)
         $yearWeek = YearWeek::of($year, $week);
         $this->assertSame($expected, (string) $yearWeek);
     }
+
+    public function testNow()
+    {
+        $timeZone = TimeZone::parse('Asia/Taipei');
+        $yearWeek = YearWeek::now($timeZone);
```
BenMorel: You should provide a Clock here, and check that the resulting YearWeek is in line with the clock!

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+
+    public function testPlusPeriod()
+    {
+
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('Thursday', (string)$zonedDateTime->plusPeriod(Period::ofWeeks(11))->getDayOfWeek());
```
BenMorel: Please assert the full string here: `assertSame('2000-...', (string) $zonedDateTime->plusPeriod(Period::ofWeeks(11)));`

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusDuration()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(56, $zonedDateTime->plusDuration(Duration::zero())->getSecond());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusYears()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(2002, $zonedDateTime->plusYears(2)->getYear());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusMonths()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(3, $zonedDateTime->plusMonths(2)->getMonth());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+
+    public function testPlusWeeks()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(2, $zonedDateTime->plusWeeks(2)->getMonth());
+        $this->assertSame(3, $zonedDateTime->plusWeeks(2)->getDay());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusDays()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(22, $zonedDateTime->plusDays(2)->getDay());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusHours()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(14, $zonedDateTime->plusHours(2)->getHour());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusMinutes()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(36, $zonedDateTime->plusMinutes(2)->getMinute());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusSeconds()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(58, $zonedDateTime->plusSeconds(2)->getSecond());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusPeriod()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('Thursday', (string)$zonedDateTime->minusPeriod(Period::ofWeeks(11))->getDayOfWeek());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusDuration()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(56, $zonedDateTime->minusDuration(Duration::zero())->getSecond());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusYears()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(1999, $zonedDateTime->minusYears(1)->getYear());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusMonths()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(12, $zonedDateTime->minusMonths(1)->getMonth());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+
+    public function testMinusWeeks()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(1, $zonedDateTime->minusWeeks(2)->getMonth());
+        $this->assertSame(6, $zonedDateTime->minusWeeks(2)->getDay());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusDays()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(18, $zonedDateTime->minusDays(2)->getDay());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusHours()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(10, $zonedDateTime->minusHours(2)->getHour());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusMinutes()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(32, $zonedDateTime->minusMinutes(2)->getMinute());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusSeconds()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(54, $zonedDateTime->minusSeconds(2)->getSecond());
```
BenMorel: Same as above

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,440 @@ public function providerCompareTo() : array
...
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '1998-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $futureZonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertTrue($futureZonedDateTime->isPast());
+    }
+
+    public function testZonedDateTimeInstanceShouldReturnString()
```
BenMorel: Please rename to `testToString()`

tests/YearMonthTest.php
peter279k: I agree that and it seems that the test method name can be short.

tests/YearWeekTest.php
peter279k: Can you give some example about this?

Thanks.

BenMorel: You can set set timestamp that `now()` will use, this way:

```php
$clock = new FixedClock(Instant::of(2000000000));
$yearWeek = YearWeek::now($timeZone, $clock);
```

tests/InstantTest.php
```diff
@@ -493,6 +564,17 @@ public function providerCompareTo() : array
         ];
     }
 
+    /**
+     * @return array
+     */
+    public function providerInclusiveBetweenCompareTo() : array
```
BenMorel: This provider can be removed as well!

tests/IntervalTest.php
```diff
@@ -12,6 +12,18 @@
  */
 class IntervalTest extends AbstractTestCase
 {
+    /**
+     * @expectedException        Brick\DateTime\DateTimeException
+     * @expectedExceptionMessage The end instant must not be before the start instant.
+     */
+    public function testEndInstantIsNotBeBeforeTheStartInstant()
```
BenMorel: Typo "BeBefore", please rename to "testEndInstantIsNotBeforeStartInstant"

tests/LocalDateTest.php
```diff
@@ -1251,6 +1252,13 @@ public function providerToString() : array
         ];
     }
 
+    public function testCheckWithNullMonthOfYear()
+    {
+        $dayOfMonth = new DayOfMonth();
+
+        $this->assertNull($dayOfMonth->check(31));
```
BenMorel: I did not mean just to move the test, but to try and trigger the case when `DayOfMonth::check()` is called with `$monthOfYear === null` when calling `LocalDate::of()`, or `LocalDate::withDay()`.

Then I realized that it's actually never called with a `null` value from anywhere (maybe it should not be nullable then...)

So sorry, the way you did before was OK for now, can you please re-introduce `DayOfMonthTest`, but please move it to a `Field` sub-namespace: `Brick\DateTime\Tests\Field\DayOfMonthTest`.

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,438 @@ public function providerCompareTo() : array
...
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(56, $zonedDateTime->getSecond());
+    }
+
+    public function testWithDate()
+    {
+        $y = (int)date('Y');
+        $m = (int)date('m');
+        $d = (int)date('d');
```
BenMorel: Please hardcode a year, month, day here. Tests must be reproducible and cannot rely on the current time.

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,438 @@ public function providerCompareTo() : array
...
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $newDate = LocalDate::of($y, $m, $d);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(date('Y-m-d'), (string)$zonedDateTime->withDate($newDate)->getDate());
```
BenMorel: Space after `(string)` please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,438 @@ public function providerCompareTo() : array
...
+
+    public function testWithTime()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $time = LocalTime::of(12, 34, 56, 987654321);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('12:34:56.987654321', (string)$zonedDateTime->withTime($time)->getTime());
```
BenMorel: Same as above, you should check the whole zoned date-time string because we need to ensure that the date is correct, as well.

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,438 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithYear()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame(2000, $zonedDateTime->withYear(2000)->getYear());
```
BenMorel: Please check the whole zoned date-time string as well:
`$this->assertSame('...', (string) $zonedDateTime->withYear(...))`

Is this on purpose that you're setting the same year as the one in the original date? Not sure which part of the code you're testing here. If not, it'd be actually more useful to test with another year (or perform both tests, same year and other year).

Same for `withMonth()`, `withDay()`, and others below.

peter279k: I do the same things for the ```withHour```, ```withMinute```, ```withSecond```, ```withNano``` and ```withFixedOffsetTimeZone```.

tests/MonthDayTest.php
```diff
@@ -376,6 +376,17 @@ public function providerWithInvalidMonthThrowsException() : array
         ];
     }
 
+    public function testWithDayWithSameDay()
+    {
+        $month = (int)date('m');
+        $day = (int)date('d');
```
BenMorel: Please hardcode month and day here.

tests/DayOfMonthTest.php
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Brick\DateTime\Tests\Field;
+
+use Brick\DateTime\Field\DayOfMonth;
+use Brick\DateTime\Tests\AbstractTestCase;
+
+/**
+ * Unit tests for class DayOfMonth.
+ */
+class DayOfMonthTest extends AbstractTestCase {
```
BenMorel: `{` on newline!

tests/DayOfMonthTest.php
```diff
@@ -0,0 +1,20 @@
+<?php
```
BenMorel: This file should be in tests/Field/DayOfMonthTest.php

tests/DayOfMonthTest.php
```diff
@@ -0,0 +1,20 @@
...
+ * Unit tests for class DayOfMonth.
+ */
+class DayOfMonthTest extends AbstractTestCase {
+    public function testCheckWithNullMonthOfYear()
+    {
+        $dayOfMonth = new DayOfMonth();
+
+        $this->assertNull($dayOfMonth->check(31));
+    }
+}
```
BenMorel: Newline at end of file, please

tests/WeekOfYearTest.php
```diff
@@ -0,0 +1,31 @@
+<?php
```
BenMorel: This file should be in tests/Field/WeekOfYearTest.php

tests/WeekOfYearTest.php
```diff
@@ -0,0 +1,31 @@
...
+
+declare(strict_types=1);
+
+namespace Brick\DateTime\Tests;
```
BenMorel: Namespace should be Brick\DateTime\Tests\Field

tests/WeekOfYearTest.php
```diff
@@ -0,0 +1,31 @@
...
+    public function testCheckShouldThrowDateTimeExceptionWithFieldNotInRange()
+    {
+        WeekOfYear::check(-1);
+    }
+
+    /**
+     * @expectedException        Brick\DateTime\DateTimeException
+     * @expectedExceptionMessage Year 2000 does not have 53 weeks
+     */
+    public function testCheckTimeShouldThrowDateTimeException()
```
BenMorel: Please rename to testCheckShouldThrowDateTimeExceptionWithNonLeapYear

tests/YearMonthTest.php
```diff
@@ -330,16 +330,38 @@ public function testWithYear()
         $this->assertYearMonthIs(2001, 5, YearMonth::of(2000, 5)->withYear(2001));
     }
 
+    public function testWithYearIsTheSameYear()
+    {
+        $fixedYear = (int)date('Y');
+        $fixedMonth = (int)date('m');
```
BenMorel: Please hardcode the year and month here. Tests need to be reproducible and cannot depend on the current date/time.

tests/YearMonthTest.php
```diff
@@ -330,16 +330,38 @@ public function testWithYear()
...
+    }
+
     public function testWithMonth()
     {
         $this->assertYearMonthIs(2000, 12, YearMonth::of(2000, 1)->withMonth(12));
     }
 
+    public function testWithMonthWithSameMonth()
+    {
+        $fixedMonth = (int)date('m');
```
BenMorel: Please hardcode the month here

tests/YearWeekTest.php
```diff
@@ -395,4 +399,31 @@ public function testToString(int $year, int $week, string $expected)
...
         $this->assertSame($expected, (string) $yearWeek);
     }
+
+    public function testNow()
+    {
+        $now = new FixedClock(Instant::of(2000000000));
+        $timeZone = TimeZone::parse('Asia/Taipei');
+        $yearWeek = YearWeek::now($timeZone, $now);
+
+        $this->assertInstanceOf(YearWeek::class, $yearWeek);
```
BenMorel: Please check the (string) representation of $yearWeek, now that you've set a fixed clock!

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+
+    public function testWithDate()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $newDate = LocalDate::of(2000, 1, 20);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->withDate($newDate));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+
+    public function testWithTime()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $time = LocalTime::of(12, 34, 56, 987654321);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.987654321-08:00[America/Los_Angeles]', (string)$zonedDateTime->withTime($time));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithYear()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->withYear(2000));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithMonth()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->withMonth(1));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithDay()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-01T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->withDay(1));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithHour()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->withHour(12));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithMinute()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:30:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->withMinute(30));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithSecond()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->withSecond(56));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithNano()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.000000056-08:00[America/Los_Angeles]', (string)$zonedDateTime->withNano(56));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithFixedOffsetTimeZone()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00', (string)$zonedDateTime->withFixedOffsetTimeZone(56));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+
+    public function testPlusPeriod()
+    {
+
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-04-06T12:34:56.123456789-07:00[America/Los_Angeles]', (string)$zonedDateTime->plusPeriod(Period::ofWeeks(11)));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+
+    public function testPlusDuration()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->plusDuration(Duration::zero()));
+    }
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+
+    public function testPlusYears()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2002-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->plusYears(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusMonths()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-03-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->plusMonths(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusWeeks()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-02-03T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->plusWeeks(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusDays()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-22T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->plusDays(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusHours()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T14:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->plusHours(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusMinutes()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:36:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->plusMinutes(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusSeconds()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:58.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->plusSeconds(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusPeriod()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('1999-11-04T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusPeriod(Period::ofWeeks(11)));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusDuration()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusDuration(Duration::zero()));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusYears()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('1999-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusYears(1));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusMonths()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('1999-12-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusMonths(1));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusWeeks()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-06T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusWeeks(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusDays()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-18T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusDays(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusHours()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T10:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusHours(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusMinutes()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:32:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusMinutes(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusSeconds()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:54.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime->minusSeconds(2));
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testToString()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string)$zonedDateTime);
```
BenMorel: Space after (string) please

tests/ZonedDateTimeTest.php
BenMorel: Question: are you setting the same date on purpose? What about setting another date, which would be more interesting to test?

Same remark for other tests below!

tests/YearMonthTest.php
```diff
@@ -332,8 +332,8 @@ public function testWithYear()
 
     public function testWithYearIsTheSameYear()
     {
-        $fixedYear = (int)date('Y');
-        $fixedMonth = (int)date('m');
+        $fixedYear = 2018;
+        $fixedMonth = 2;
```
BenMorel: Please call them just $year and $month!

tests/YearMonthTest.php
```diff
@@ -347,7 +347,7 @@ public function testWithMonth()
 
     public function testWithMonthWithSameMonth()
     {
-        $fixedMonth = (int)date('m');
+        $fixedMonth = 2;
```
BenMorel: Should be called simply $month

tests/YearWeekTest.php
```diff
@@ -406,7 +406,7 @@ public function testNow()
         $timeZone = TimeZone::parse('Asia/Taipei');
         $yearWeek = YearWeek::now($timeZone, $now);
 
-        $this->assertInstanceOf(YearWeek::class, $yearWeek);
```
BenMorel: You can keep the instanceof check

tests/Field/WeekOfYearTest.php
```diff
@@ -0,0 +1,32 @@
...
+    public function testCheckShouldThrowDateTimeExceptionWithFieldNotInRange()
+    {
+        WeekOfYear::check(-1);
+    }
+
+    /**
+     * @expectedException        Brick\DateTime\DateTimeException
+     * @expectedExceptionMessage Year 2000 does not have 53 weeks
+     */
+    public function testCheckShouldThrowDateTimeExceptionWithNonLeapYear()
```
BenMorel: Sorry, my mistake, it's not a question of leap year here.
The name should be testCheckShouldThrowDateTimeExceptionWith52WeekYear()

tests/InstantTest.php
```diff
@@ -553,4 +624,13 @@ public function providerToString() : array
             [2000000000, 0, '2033-05-18T03:33:20Z'],
         ];
     }
+
+    public function testAtTimeZone()
+    {
+        $timeZone = TimeZone::parse('UTC');
```
BenMorel: Can be `TimeZone::utc()`

tests/YearWeekTest.php
```diff
@@ -395,4 +399,31 @@ public function testToString(int $year, int $week, string $expected)
...
+
+        $this->assertSame('2033-W20', (string) $yearWeek);
+    }
+
+    public function testGetFirstDay()
+    {
+        $yearWeek = YearWeek::of(2015, 1);
+        $result = $yearWeek->getFirstDay();
+
+        $this->assertInstanceOf(LocalDate::class, $result);
```
BenMorel: This line can be removed, `assertLocalDate()` is typed and will throw a TypeError if it is not a LocalDate.

tests/YearWeekTest.php
```diff
@@ -395,4 +399,31 @@ public function testToString(int $year, int $week, string $expected)
...
+        $this->assertLocalDateIs(2014, 12, 29, $result);
+    }
+
+    public function testGetLastDay()
+    {
+        $yearWeek = YearWeek::of(2015, 1);
+        $result = $yearWeek->getLastDay();
+
+        $this->assertInstanceOf(LocalDate::class, $result);
```
BenMorel: Same, can be safely removed.

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +486,435 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testIsPast()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '1998-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $futureZonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertTrue($futureZonedDateTime->isPast());
```
BenMorel: You're implicitly testing with SystemClock (the current time), which is not a good idea in tests.
It would be nice to provide a FixedClock to `isFuture()` and `isPast()`.

Also, it would be nice to have a data provider with a few tests in there: the provider can return [$clockTimestamp, $dateTime, $isFuture], with a couple timestamps, and at least 2 datetimes each (one past, one future).

You can actually use the same data provider for both test methods:
The first one will do `assertSame($isFuture, $zonedDateTime->isFuture())`, while the second one will do `assertSame(! $isFuture, $zonedDateTime->isPast())`.

tests/Field/DayOfMonthTest.php
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Unit tests for class DayOfMonth.
+ */
+class DayOfMonthTest extends AbstractTestCase
+{
+    public function testCheckWithNullMonthOfYear()
+    {
+        $dayOfMonth = new DayOfMonth();
+
+        $this->assertNull($dayOfMonth->check(31));
```
BenMorel: I'm not comfortable with assertNull() for a void method. To remove the assertion and avoid a "risky" test, maybe you can give a try to [@doesNotPerformAssertions](https://github.com/sebastianbergmann/phpunit/issues/2484#issuecomment-278233214)?

peter279k: It sounds good because it's the void method and not return the ```null``` currently.

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +487,448 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithFixedOffsetTimeZone()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-10T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-10T12:34:56.123456789-08:00', (string) $zonedDateTime->withFixedOffsetTimeZone(56));
```
BenMorel: withFixedOffsetTimeZone() does not take any parameters!

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +487,448 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusPeriod()
+    {
+
```
BenMorel: Extra line

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +487,448 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testPlusDuration()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string) $zonedDateTime->plusDuration(Duration::zero()));
```
BenMorel: Could you please use another Duration than zero?

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +487,448 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testMinusDuration()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-01-20T12:34:56.123456789-08:00[America/Los_Angeles]', (string) $zonedDateTime->minusDuration(Duration::zero()));
```
BenMorel: Same, a non-zero Duration would be more useful!

tests/DateTimeParseResultTest.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Brick\DateTime\Tests;
```
BenMorel: This should be in namespace `Brick\DateTime\Tests\Parser`, and located in a `tests/Parser` directory

tests/PatternParserBuilderTest.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Brick\DateTime\Tests;
+
+use Brick\DateTime\Parser\PatternParserBuilder;
```
BenMorel: This should be in namespace `Brick\DateTime\Tests\Parser`, and located in a `tests/Parser` directory

tests/PatternParserBuilderTest.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Brick\DateTime\Tests;
```
BenMorel: This should be in namespace `Brick\DateTime\Tests\Parser`, and located in a `tests/Parser` directory

tests/PatternParserBuilderTest.php
```diff
@@ -0,0 +1,44 @@
...
+
+use Brick\DateTime\Parser\PatternParserBuilder;
+
+/**
+ * Unit tests for class IsoParsers.
```
BenMorel: for class PatternParserBuilder.

tests/DateTimeParseResultTest.php
```diff
@@ -0,0 +1,23 @@
...
+
+use Brick\DateTime\Parser\DateTimeParseResult;
+
+/**
+ * Unit tests for class DateTimeParseResult.
+ */
+class DateTimeParseResultTest extends AbstractTestCase
+{
+    /**
+     * @expectedException        Brick\DateTime\Parser\DateTimeParseException
```
BenMorel: Please add leading `\` before exception class

tests/Field/WeekOfYearTest.php
```diff
@@ -0,0 +1,32 @@
...
+use Brick\DateTime\Field\WeekOfYear;
+use Brick\DateTime\Tests\AbstractTestCase;
+
+/**
+ * Unit tests for class WeekOfYear.
+ */
+class WeekOfYearTest extends AbstractTestCase
+{
+    /**
+     * @expectedException        Brick\DateTime\DateTimeException
```
BenMorel: Please add leading `\` before exception class

tests/Field/WeekOfYearTest.php
```diff
@@ -0,0 +1,32 @@
...
+     * @expectedExceptionMessage Invalid week-of-year: -1 is not in the range 1 to 53.
+     */
+    public function testCheckShouldThrowDateTimeExceptionWithFieldNotInRange()
+    {
+        WeekOfYear::check(-1);
+    }
+
+    /**
+     * @expectedException        Brick\DateTime\DateTimeException
```
BenMorel: Please add leading `\` before exception class

tests/IntervalTest.php
```diff
@@ -12,6 +12,18 @@
  */
 class IntervalTest extends AbstractTestCase
 {
+    /**
+     * @expectedException        Brick\DateTime\DateTimeException
```
BenMorel: Please add leading `\` before exception class

tests/PatternParserBuilderTest.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        $patternParserBuilder = new PatternParserBuilder();
+        $patternParserBuilder->endOptional();
+    }
+
+    /**
+     * @expectedException        \RuntimeException
+     * @expectedExceptionMessage Cannot call endGroup() without a call to startGroup() first.
+     */
+    public function testEndGroupShouldThrowRunTimeException()
```
BenMorel: RunTime => Runtime

tests/YearMonthTest.php
```diff
@@ -330,16 +330,38 @@ public function testWithYear()
         $this->assertYearMonthIs(2001, 5, YearMonth::of(2000, 5)->withYear(2001));
     }
 
+    public function testWithYearIsTheSameYear()
```
BenMorel: Please rename to testWithYearWithSameYear

tests/YearMonthTest.php
```diff
@@ -330,16 +330,38 @@ public function testWithYear()
...
+    {
+        $year = 2018;
+        $month = 2;
+        $yearMonth = YearMonth::of($year, $month);
+
+        $this->assertInstanceOf(YearMonth::class, $yearMonth->withYear($year));
```
BenMorel: This line is not needed, assertYearMonthIs() will throw a TypeError if it's not an instance of YearMonth.

tests/YearMonthTest.php
```diff
@@ -330,16 +330,38 @@ public function testWithYear()
...
+
+        $this->assertYearMonthIs(2000, $month, YearMonth::of(2000, $month)->withMonth($month));
+    }
+
     public function testGetFirstDay()
     {
         $this->assertLocalDateIs(2023, 10, 1, YearMonth::of(2023, 10)->getFirstDay());
     }
 
+    public function testGetLastDay()
```
BenMorel: Actually there was already a test for getLastDay(), you can remove this method! See below.

tests/YearWeekTest.php
```diff
@@ -395,4 +399,30 @@ public function testToString(int $year, int $week, string $expected)
...
     }
+
+    public function testNow()
+    {
+        $now = new FixedClock(Instant::of(2000000000));
+        $timeZone = TimeZone::parse('Asia/Taipei');
+        $yearWeek = YearWeek::now($timeZone, $now);
+
+        $this->assertSame('2033-W20', (string) $yearWeek);
+        $this->assertInstanceOf(YearWeek::class, $yearWeek);
```
BenMorel: Can you use `assertYearWeekIs()` instead of these 2 lines?

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +487,448 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testGetDayOfWeek()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-01-20T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertDayOfWeekIs(4, $zonedDateTime->getDayOfWeek());
```
BenMorel: Please use the `DayOfWeek::THURSDAY` constant instead of `4` for readability

tests/ZonedDateTimeTest.php
```diff
@@ -482,4 +487,448 @@ public function providerCompareTo() : array
...
+    }
+
+    public function testWithSecond()
+    {
+        $timeZone = TimeZone::parse('America/Los_Angeles');
+        $localDateTime = '2000-03-25T12:34:56.123456789';
+        $localDateTime = LocalDateTime::parse($localDateTime);
+        $zonedDateTime = ZonedDateTime::of($localDateTime, $timeZone);
+
+        $this->assertSame('2000-03-25T12:34:56.123456789-08:00[America/Los_Angeles]', (string) $zonedDateTime->withSecond(56));
```
BenMorel: Can you use another second value in `withSecond()`? It's the same as in the original date-time!

tests/Field/DayOfMonthTest.php
```diff
@@ -0,0 +1,23 @@
...
+
+namespace Brick\DateTime\Tests\Field;
+
+use Brick\DateTime\Field\DayOfMonth;
+use Brick\DateTime\Tests\AbstractTestCase;
+
+/**
+ * Unit tests for class DayOfMonth.
+ *
+ * @doesNotPerformAssertions
```
BenMorel: Does it work with the annotation on the method rather than on the class?
It would be better if we add other methods in the future.

tests/Field/DayOfMonthTest.php
```diff
@@ -0,0 +1,23 @@
...
+ */
+class DayOfMonthTest extends AbstractTestCase
+{
+    public function testCheckWithNullMonthOfYear()
+    {
+        $dayOfMonth = new DayOfMonth();
+
+        $dayOfMonth->check(31);
```
BenMorel: `check()` is a static method, it should be `DayOfMonth::check(31);` and the first line can be removed.

tests/Field/DayOfMonthTest.php
peter279k: Yes. I set the wrong place for the annotation.

BenMorel:
I think you forgot to reintroduce DateTimeParseResultTest and PatternParserBuilderTest, they're missing!

BenMorel:
Looks good now 👍 Thanks!

PR #25, opened 2020-06-30
Fix method plusDuration to take into account the current time.

JodyLognoul:
The method `plusDuration` of class LocalDateTime must take into account the current time to add the duration.


BenMorel:
Thank you! Released as [0.2.2](https://github.com/brick/date-time/releases/tag/0.2.2) 👍

PR #26, opened 2020-08-02
Add static analysis using Psalm, and fix reported issues

bdsl:
This is using the second strictest set of checks in Psalm default
configs. It might be interesting to get the library working on
the strictest level, which would mean resolving all the issues with
mixed types

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ before_script:
 
 script:
   - mkdir -p build/logs
+  - vendor/bin/psalm.phar
   - vendor/bin/phpunit --coverage-clover build/logs/clover.xml
 
 after_script:
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "require-dev": {
         "phpunit/phpunit": "^7.5.15",
         "php-coveralls/php-coveralls": "^2.2",
-        "psalm/phar": "^3.13"
+        "psalm/phar": "^4.1.1"
     },
     "suggest": {
         "ext-timezonedb": "This PECL extension provides up-to-date timezone information"
--- a/composer.json
+++ b/composer.json
@@ -9,12 +9,12 @@
     ],
     "license": "MIT",
     "require": {
-        "php": "^7.1|^8.0",
-        "psalm/phar": "^3.13"
+        "php": "^7.1|^8.0"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.5.15",
-        "php-coveralls/php-coveralls": "^2.2"
+        "php-coveralls/php-coveralls": "^2.2",
+        "psalm/phar": "^3.13"
     },
     "suggest": {
         "ext-timezonedb": "This PECL extension provides up-to-date timezone information"
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,8 @@
     ],
     "license": "MIT",
     "require": {
-        "php": "^7.1|^8.0"
+        "php": "^7.1|^8.0",
+        "psalm/phar": "^3.13"
     },
     "require-dev": {
         "phpunit/phpunit": "^7.5.15",
--- a/psalm.xml
+++ b/psalm.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="2"
+    resolveFromConfigFile="true"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://getpsalm.org/schema/config"
+    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+>
+    <projectFiles>
+        <directory name="src" />
+        <ignoreFiles>
+            <directory name="vendor" />
+        </ignoreFiles>
+    </projectFiles>
+</psalm>
--- a/src/Duration.php
+++ b/src/Duration.php
@@ -165,7 +165,6 @@ final class Duration implements \JsonSerializable
             $seconds--;
         }
 
-
         return new Duration($seconds, $nanoseconds);
     }
 
--- a/src/Duration.php
+++ b/src/Duration.php
@@ -158,11 +158,13 @@ final class Duration implements \JsonSerializable
     {
         $nanoseconds = $nanoAdjustment % LocalTime::NANOS_PER_SECOND;
         $seconds += ($nanoAdjustment - $nanoseconds) / LocalTime::NANOS_PER_SECOND;
+        \assert(\is_int($seconds));
 
         if ($nanoseconds < 0) {
             $nanoseconds += LocalTime::NANOS_PER_SECOND;
             $seconds--;
         }
+
 
         return new Duration($seconds, $nanoseconds);
     }
--- a/src/LocalDateRange.php
+++ b/src/LocalDateRange.php
@@ -201,7 +201,7 @@ final class LocalDateRange implements \IteratorAggregate, \Countable, \JsonSeria
     /**
      * Returns an iterator for all the dates contained in this range.
      *
-     * @return LocalDate[]
+     * @return \Generator<LocalDate>
      */
     public function getIterator() : \Generator
     {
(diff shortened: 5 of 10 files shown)


psalm.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0"?>
+<psalm
+    errorLevel="2"
+    resolveFromConfigFile="true"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xmlns="https://getpsalm.org/schema/config"
+    xsi:schemaLocation="https://getpsalm.org/schema/config vendor/vimeo/psalm/config.xsd"
+>
+    <projectFiles>
+        <directory name="src" />
```
bdsl, line 10: Might want to add `<directory name="tests" />` but that would mean more issues to be fixed, suppressed, or baselined.

composer.json
```diff
@@ -9,7 +9,8 @@
     ],
     "license": "MIT",
     "require": {
-        "php": "^7.1|^8.0"
+        "php": "^7.1|^8.0",
+        "psalm/phar": "^3.13"
```
jdreesen, line 13: Psalm should be a `dev` dependency.

bdsl: Thanks, you're right of course. Having it in `require` would cause conflicts with projects that use this library and require a different version of `psalm/phar`. Fixed in b4f457b

BenMorel:
Hi, thanks for your PR! I'm waiting for https://github.com/brick/math/pull/43 to be merged, then I'll come to you 👍

bdsl:
Hi Ben. I hadn't seen brick/math#43 . Is there are a dependency between that and this PR? Or you want to learn from that one before you decide what what to do here?

Let me know if you'd like me to rebase this on the latest master.

If you think it would be useful I'd be up for adding more checking tools like PHPStan or perhaps Infection, and maybe increasing the Psalm strictness level and making it cover tests as well as source.

src/LocalDateRange.php
```diff
@@ -201,7 +201,7 @@ public function getIntersectionWith(LocalDateRange $that) : LocalDateRange
     /**
      * Returns an iterator for all the dates contained in this range.
      *
-     * @return LocalDate[]
+     * @return \Generator<LocalDate>
```
bdsl, line 204:
```suggestion
     * @return \Iterator<LocalDate>
```
Let's not make the contract more specific than it needs to be.

src/YearMonthRange.php
```diff
@@ -159,7 +159,7 @@ public function contains(YearMonth $yearMonth) : bool
     /**
      * Returns an iterator for all the year-months contained in this range.
      *
-     * @return YearMonth[]
+     * @return \Generator<YearMonth>
```
bdsl, line 162:
```suggestion
     * @return \Iterator<YearMonth>
```

bdsl:
I went ahead and rebased on latest master ( 3d29c70a35075 ) to make sure it still passes.

BenMorel:
> Or you want to learn from that one before you decide what what to do here?

Exactly.

> If you think it would be useful I'd be up for adding more checking tools like PHPStan or perhaps Infection, and maybe increasing the Psalm strictness level and making it cover tests as well as source.

+1 for maximum strictness, if that doesn’t affect legitimate code (example: in the other PR, adding a method was suggested, just to make Psalm happy - I’m not OK with that).

I’m still not sure if I want to use both Psalm *and* phpstan (what about phan?), I’ll be happy to hear your thoughts about this.

I’m also willing to add a coding style, with phpcs and/or php-cs-fixer. Infection, why not. But I’ll be using brick/math (my most popular library) as a testing ground before deploying this to all my libraries, to try and keep them consistent with each other!

I appreciate to have 2 separate contributions for code quality though, and would be grateful if you can participate to https://github.com/brick/math/pull/43!

bdsl:
Ok, I'll try and find time to take a good look at brick/math#43 the other soon and I can try and answer some of your questions about it if @alexeyshockov hasn't answered them first, and maybe also join the discussion about the coding style tool over at brick/math.

About phpstan, I think it has several types of issues it can catch that Psalm doesn't, and vice versa, so you find the most possible issues by using both. I only have experience with Psalm personally.

bdsl:
@BenMorel Do you want to take another look at this now that https://github.com/brick/math/pull/43#issuecomment-719548336 is closed and Psalm is working in brick/math ?

src/ZonedDateTime.php
```diff
@@ -232,6 +232,12 @@ public static function fromDateTime(\DateTimeInterface $dateTime) : ZonedDateTim
 
         $dateTimeZone = $dateTime->getTimezone();
 
+        /**
+         * @psalm-suppress TypeDoesNotContainType
+         *
+         * Looks like Psalm might have a wrong return type for \DateTimeInterface::getTimeZone - it should
+         * be \DateTimeZone|false . Psalm seems to think it's just \DateTimeZone
+         */
```
BenMorel, line 240: Reported here: https://github.com/vimeo/psalm/issues/4515
Please add a link in the comment!

src/Duration.php
```diff
@@ -158,12 +158,14 @@ public static function ofSeconds(int $seconds, int $nanoAdjustment = 0) : Durati
...
         $nanoseconds = $nanoAdjustment % LocalTime::NANOS_PER_SECOND;
         $seconds += ($nanoAdjustment - $nanoseconds) / LocalTime::NANOS_PER_SECOND;
+        \assert(\is_int($seconds));
 
         if ($nanoseconds < 0) {
             $nanoseconds += LocalTime::NANOS_PER_SECOND;
             $seconds--;
         }
 
+
```
BenMorel, line 168: Please remove this extra line

composer.json
```diff
@@ -13,7 +13,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^7.5.15",
-        "php-coveralls/php-coveralls": "^2.2"
+        "php-coveralls/php-coveralls": "^2.2",
+        "psalm/phar": "^3.13"
```
BenMorel, line 17: Could you please upgrade to Psalm ^4.1?

.travis.yml
```diff
@@ -11,6 +11,7 @@ before_script:
 
 script:
   - mkdir -p build/logs
+  - vendor/bin/psalm.phar
```
BenMorel, line 14: Instead of running Psalm with every build, it would be nice to create a separate job for it.
You can take inspiration from brick/math: https://github.com/brick/math/blob/master/.travis.yml

psalm.xml
BenMorel: Let's do `src` only for now, you can still propose another PR for `tests` later on!

BenMorel:
@bdsl Please check my comments, and it will be good to go! :+1:

bdsl:
@BenMorel Thanks, I've applied all your suggestions apart from the one about travis. I was going to consider that next, but maybe Travis is having problems - the job for building 0777a78 seems to have been queued for several minutes. I'm not sure how long it should normally take. I'll take another look later and see if it's run.

composer.json
bdsl: Done

src/Duration.php
bdsl: done

src/ZonedDateTime.php
bdsl: Done

.travis.yml
bdsl: I'm not clear on what the benefit of the separate job would be at this stage. Psalm only seems to take 1.15 seconds to run, so the overheads of setting up a separate job probably wouldn't be justified by the potential savings of running it in parallel with other things. I'm also not very familiar with Travis.

BenMorel: You have a point here, though it's more of a matter of proper job separation, so that when a job fails, you know that it's because of a failed test and not because of a Psalm issue. This also gives the ability to use Psalm as one of the "smoke tests" that prevent other tests from running if it doesn't pass. Let's merge it this way for now, I'll come back to this later.

BenMorel:
Thanks, @bdsl!

PR #31, opened 2021-02-19
add toDatePeriod method for LocalDateRange

morrislaptop:
Adds a `toDatePeriod` to the `LocalDateRange` object which returns a native \DatePeriod PHP object. 

I opted to do end of day (instead of midnight next day) to be inline with the inclusive nature of LocalDateRange.

Note: This is not `bricks/date-time` Period class - I hope this doesn't create confusion.

### DIFF
--- a/src/LocalDateRange.php
+++ b/src/LocalDateRange.php
@@ -234,7 +234,8 @@ final class LocalDateRange implements \IteratorAggregate, \Countable, \JsonSeria
     /**
      * Converts this LocalDateRange to a native DatePeriod object.
      *
-     * The result is a DateTime with time 00:00 in the UTC time-zone.
+     * The result is a DatePeriod->start with time 00:00 and a DatePeriod->end 
+     * with time 23:59:59.999999 in the UTC time-zone.
      *
      * @return \DatePeriod
      */
--- a/src/LocalDateRange.php
+++ b/src/LocalDateRange.php
@@ -8,6 +8,7 @@ use Brick\DateTime\Parser\DateTimeParseException;
 use Brick\DateTime\Parser\DateTimeParser;
 use Brick\DateTime\Parser\DateTimeParseResult;
 use Brick\DateTime\Parser\IsoParsers;
+use DatePeriod;
 
 /**
  * Represents an inclusive range of local dates.
@@ -228,6 +229,22 @@ final class LocalDateRange implements \IteratorAggregate, \Countable, \JsonSeria
     public function jsonSerialize() : string
     {
         return (string) $this;
+    }
+
+    /**
+     * Converts this LocalDateRange to a native DatePeriod object.
+     *
+     * The result is a DateTime with time 00:00 in the UTC time-zone.
+     *
+     * @return \DatePeriod
+     */
+    public function toDatePeriod() : \DatePeriod
+    {
+        $start = $this->getStart()->atTime(LocalTime::midnight())->toDateTime();
+        $end = $this->getEnd()->atTime(LocalTime::max())->toDateTime();
+        $interval = new \DateInterval('P1D');
+
+        return new \DatePeriod($start, $interval, $end);
     }
 
     /**
--- a/tests/LocalDateRangeTest.php
+++ b/tests/LocalDateRangeTest.php
@@ -222,18 +222,19 @@ class LocalDateRangeTest extends AbstractTestCase
     /**
      * @dataProvider providerToDatePeriod
      *
-     * @param string $dateTime The date-time string that will be parse()d by LocalDate.
-     * @param string $expected The expected output from the native DateTime object.
+     * @param string $range The date-time string that will be parse()d by LocalDateRange.
+     * @param string $expectedStart The expected output from the native DateTime object.
+     * @param string $expectedEnd The expected output from the native DateTime object.
      */
-    public function testToDatePeriod($a, $expectedStart, $expectedEnd)
+    public function testToDatePeriod(string $range, string $expectedStart, string $expectedEnd)
     {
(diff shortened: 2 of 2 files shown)


BenMorel:
Hi, this can be a great addition indeed.

Are you sure we shouldn't use the next day though, for the end date?

Otherwise the last date is not included when you iterate the `DatePeriod` object, which makes it behave differently from `LocalDateRange` in this regard: https://3v4l.org/5Zggu

src/LocalDateRange.php
```diff
@@ -230,6 +231,22 @@ public function jsonSerialize() : string
         return (string) $this;
     }
 
+    /**
+     * Converts this LocalDateRange to a native DatePeriod object.
+     *
+     * The result is a DateTime with time 00:00 in the UTC time-zone.
```
BenMorel, line 237: Is this copy/pasted from elsewhere?

morrislaptop: Yep! Fixed up.

morrislaptop:
> Hi, this can be a great addition indeed.
> 
> Are you sure we shouldn't use the next day though, for the end date?
> 
> Otherwise the last date is not included when you iterate the `DatePeriod` object, which makes it behave differently from `LocalDateRange` in this regard: https://3v4l.org/5Zggu

Both the LocalDateRange and DatePeriod exclude the last day in the range - I've added some assertions to verify this.

morrislaptop:
Ah yep, got you now. I feel like the essence of the library is inclusive, whereas \DatePeriod is exclusive.  So there's a mismatch of ideas here. 

What do you think users would expect? 

In my scenario I just want to grab the end of the range in a DateTime context and stick it in a database.

Perhaps it might be better to do `toEndDateTime()` and `toStartDateTime()` which respect the inclusive nature of this library.

`toDatePeriod` would be exclusive which would respect the exclusive nature of the native object.

BenMorel:
Could this be configured with a flag?

```php
public function toDatePeriod(bool $nextDay)
```

I'm afraid there isn't an ideal solution anyway, as you said there's a mismatch of ideas between brick's `LocalDateRange` and `DatePeriod`.

morrislaptop:
Actually, I don't see the issue here. Both objects produce the same result when iterating.

The example you provided had the end date at the start of the day, whereas this method sets the end date at the end of the day.

The native DatePeriod will include the last day, even if it's one second past the start of the day.

This is what the latest version of the tests ensure.

E.g. https://3v4l.org/WSEnO

I think it's good to go as is :)

morrislaptop:
Hey @BenMorel , do you think you'll be merging this? I've implemented something similar over at https://github.com/spatie/period/pull/81

tests/LocalDateRangeTest.php
```diff
@@ -219,6 +219,42 @@ public function testToString()
         ));
     }
 
+    /**
+     * @dataProvider providerToDatePeriod
+     *
+     * @param string $dateTime The date-time string that will be parse()d by LocalDate.
+     * @param string $expected The expected output from the native DateTime object.
```
BenMorel, line 226: Copy/paste! Please clean up.

tests/LocalDateRangeTest.php
```diff
@@ -219,6 +219,42 @@ public function testToString()
...
+     */
+    public function testToDatePeriod($a, $expectedStart, $expectedEnd)
```
BenMorel, line 228: Please add `string` types, and rename `$a` to `$dateRange`.

tests/LocalDateRangeTest.php
```diff
@@ -219,6 +219,42 @@ public function testToString()
...
+
+        $this->assertSame(iterator_count($period), $range->count());
+        $this->assertSame($expectedStart, $period->start->format('Y-m-d\TH:i:s.uO'));
+        $this->assertSame($expectedEnd, $period->end->format('Y-m-d\TH:i:s.uO'));
+    }
+
+    /**
+     * @return array
+     */
+    public function providerToDatePeriod()
```
BenMorel, line 249: please add `array` return type.

BenMorel:
> The native DatePeriod will include the last day, even if it's one second past the start of the day.

I missed this detail, thanks. One more gotcha with native DateTime classes ;-)

Please fix the few points I mentioned and it should be good to go.

tests/LocalDateRangeTest.php
```diff
@@ -219,6 +219,42 @@ public function testToString()
...
+     */
+    public function testToDatePeriod($a, $expectedStart, $expectedEnd)
+    {
+        $range = LocalDateRange::parse($a);
+
+        $period = $range->toDatePeriod();
+
+        $range_arr = iterator_to_array($range);
+        $period_arr = iterator_to_array($period);
+        $zip = array_map(null, $range_arr, $period_arr);
```
BenMorel, line 236: Please don't use snake case for consistency with the rest of the project. `$rangeArray` and `$periodArray` would do fine.

tests/LocalDateRangeTest.php
morrislaptop: Done :)

tests/LocalDateRangeTest.php
morrislaptop: Done :)

tests/LocalDateRangeTest.php
morrislaptop: Done :)

tests/LocalDateRangeTest.php
morrislaptop: Done :)

BenMorel:
Thank you. Released as [0.2.3](https://github.com/brick/date-time/releases/tag/0.2.3)!

Issue #24, opened 2020-04-05
UtcDateTime


solodkiy:
I have a goal to change all of internal date time operations in my project to UTC dates and now looking for the simplest solution. I think that one of approach is to have class which represent such date concept like UTC date time and use it in all internal code and in db. Convert to LocalDateTime and ZonedDateTime only when it relay necessary. 

In this lib we have:

- ZonedDateTime - real point in time, and can be used, but we always need to think about time zones (is it set correctly to this object?) and also you need always convert it to UTC before saving in DB if you want UTC time there which is not very convenient and can lead to mistakes.
- LocalDateTime - can represent UTC time, but also can not represent. Looks like it should be used for interactions with other systems (api which expect date in Europe\Moscow timezone) and to render for user.
- Instant - looks very similar for my goal, but has not very convenient interface.

What do you think about adding new UtcDateTime class?
UtcDateTime is represent point absolute point in time (sounds similar to Instant, yeah)
And have all methods of LocalDateTime and some more like easy convert to ZonedDateTime(UTC) or Instant.
Or maybe we should just add a simple methods to convert ZonedDateTime to and from UTC LocalDateTime? Something like $zoned->getUtcLocal();

What to you think about it?

BenMorel:
Hi, what you're calling `UtcDateTime` really is `ZonedDateTime` with a UTC timezone; I can see no advantage in introducing another class.

- `LocalDateTime` is useful when you need to represent a date, without having to think about the timezone. It cannot represent a point in time without a timezone.
- `Instant` is really just an abstraction around a timestamp. It represents a point in time, but cannot represent a date and time without a timezone.
- `ZonedDateTime` represents both a date and a time, *and* a point in time. It can be used with the UTC timezone to do what you want.

You can convert between `LocalDateTime` and `ZonedDateTime` easily:

```php
$localDateTime = LocalDateTime::parse('2020-04-01T00:00:00');

$utcDateTime = $localDateTime->atTimeZone(TimeZone::utc()); // ZonedDateTime
$localDateTime = $utcDateTime->getDateTime(); // LocalDateTime
```

Let me know if that solves it for you!

BenMorel:
Closing due to lack of feedback. Feel free to comment if you think this should be reopened.

solodkiy:
[...]

PR #36, opened 2021-05-30
Add UtcDateTime class

solodkiy:


BenMorel:
Hi, what's the point of this class? Just avoiding passing `TimeZone::utc()`?

solodkiy:
This request if part of https://github.com/brick/date-time/issues/24 discussion.
The main point is with it you could use it for properties, arguments and return types.

PR #38, opened 2021-09-21
Fix return type of TimeZoneRegion::parse()

adrianguenter:
`TimeZoneRegion::parse(...)` should return `TimeZoneRegion` and not `TimeZone`, matching the return type of the proxied `from(...)` method.


BenMorel:
Good point. Thank you!

Issue #10, opened 2018-10-27
Resolve API inconsistencies regarding scalars versus objects


joshdifabio:
There are some inconsistencies in the API, e.g.

```php
class LocalDate
{
    public function getDayOfWeek(): DayOfWeek;

    public function getMonth(): int;
}
```

I understand why we've got `getDay(): int` and `getYear(): int` -- those values are naturally integer, and java.time makes the same judgement -- but what's the reason for treating `getDayOfWeek()` differently to `getMonth()`? These are naturally both enums.

BenMorel:
You're right that these are inconsistencies in the API. I found the decisions quite hard to make at the time I wrote these.

As I see it now, I'd say that these should all return integers. An object can be explicitly requested by calling, for example, `DayOfWeek::of($date->getDayOfWeek())` if needed.

What do you think?

And do you see other inconsistencies?

joshdifabio:
>  I found the decisions quite hard to make at the time I wrote these.

I can imagine! This really is an excellent package though and it's clear that a lot of time, thought and skill has gone into it -- thank you so much for taking the time to create it and open source it.

Regarding the matter at hand, the first thing I'd say is that we can't possibly replicate the collective brainpower and empirical evidence that went into designing JSR 310 and so, where practical, I would suggest simply copying JSR 310's API. I think we are unlikely to make better subjective judgements than the original authors made, particularly because we have far less experience with these APIs than the authors of JSR 310 had (JSR 310 is primarily a formalisation of an API which was used by thousands of developers over a number of years before JSR 310 came along, as you likely already know).

Regarding java.time's API, values which are Enums are primarily passed around as Enums (`Month` and `DayOfWeek`), not scalars. So, all of the `getMonth()` and `getDayOfWeek()` methods return Enums, not ints. I would simply replicate this here. `getYear()` and `getDayOfMonth()` are different because these are not Enums.

In terms of practical benefits, and my personal opinion, passing integers around means we have to constantly validate values to check whether they are valid months/weekdays, which negates one of the benefits of this package. Value objects and enums can eliminate a lot of boilerplate. Therefore, the enums are more useful than bare integers, and so I would primarily use them where they're available. Furthermore, months and weekdays are not naturally _numeric_. We don't refer to month 7 or weekday 3, we say _July_ and _Tuesday_, whereas we do refer to years and days of the month in numeric terms. 

However, I would very much put my opinions as secondary to those of the original authors of JSR 310, which just happen to be the same in this instance.

[...]

PR #48, opened 2022-04-10
Add LocalDateRange::toDuration, YearMonthRange::toDuration methods

solodkiy:


solodkiy:
Also we could add this method to YearMonthRange (looks reasonable), and Year, YearMonth, YearWeek (not so sure).

tests/LocalDateRangeTest.php
```diff
@@ -172,6 +173,14 @@ public function testCount(string $range, int $count): void
         $this->assertCount($count, LocalDateRange::parse($range));
     }
 
+    /**
+     * @dataProvider providerCount
+     */
+    public function testGetDuration(string $range, int $daysCount): void
+    {
+        $this->assertEquals(Duration::ofDays($daysCount), LocalDateRange::parse($range)->getDuration());
```
BenMorel, line 181: Could you please use `assertDurationIs()` to avoid using equality on objects?

BenMorel:
Apart from my comment, LGTM :+1:

Could you please consider adding `LocalDateRange::getPeriod()` as well in another PR? Although that one is less straightforward, as we have 2 ways to convert it to a period: calculating x years, y months & z days, or just 0 years, 0 months, and n days. Maybe we should even offer both ways?

`YearMonthRange::getPeriod()` would make sense, but not `getDuration()` as you cannot convert it to a number of days.

As for `Year`, `YearMonth`, `YearWeek`, they're not ranges, so I don't see how this could translate to a `Duration` or `Period`?

solodkiy:
> YearMonthRange::getPeriod() would make sense, but not getDuration() as you cannot convert it to a number of days.

Why not? YearMonthRange is inclusive. It include full first month, everything in between and full last month.
So I think YearMonthRange include all days between first day of first range month and last day of last range month (include corner days).
This approach allow us to convert YearMonthRange to LocalDateRange and to Duration.

solodkiy:
Year is not really a range, but also could represent a whole year from first day to last.

BenMorel:
I see. Let's start with this one and discuss the others in another PR.

Also, Java's threeten-extra has a [toPeriod](https://www.threeten.org/threeten-extra/apidocs/org.threeten.extra/org/threeten/extra/LocalDateRange.html#toPeriod()) method, that we may replicate here; maybe we should name this one `toDuration()` for consistency then?

solodkiy:
>maybe we should name this one toDuration() for consistency then?
I am not sure we should fight for consistency with Java here.

Can you describe main logic of choosing between `to` and `get` methods?

BenMorel:
I have no idea, but this project started as port of Java's ThreeTen API, so whenever we can keep the same API, I think we should do it for two reasons:

- there are many people behind this initiative, so they may have had this discussion before, and may have come to this consensus for a reason; to quote another contributor's [comment](https://github.com/brick/date-time/issues/10#issuecomment-433855307):
    > (...) we can't possibly replicate the collective brainpower and empirical evidence that went into designing JSR 310 and so, where practical, I would suggest simply copying JSR 310's API
- this can make this project less confusing for people coming from a Java background

As for the logic behind this, what about:

- `get*()` for fields that are *part* of the object (`getYear()`, `getMonth()`, ...)
- `to*()` for objects that are not part of it, but can be derived from it? (`toDuration()`)

Though I'm sure you can find a counter-example in the codebase.

solodkiy:
>Though I'm sure you can find a counter-example in the codebase.

I will look for it, maybe it good point to do some rename for consistency then.

BenMorel:
I finally chose `toPeriod()` in #51 for consistency with threeten-extra, so please let's use `toDuration()` here for consistency.

> > Though I'm sure you can find a counter-example in the codebase.
> 
> I will look for it, maybe it good point to do some rename for consistency then.

Thank you, that's be great.

solodkiy:
@BenMorel I renamed method, and also added YearMonthRange::toDuration

src/LocalDateRange.php
```diff
@@ -228,6 +228,16 @@ public function count() : int
...
     }
 
+    /**
+     * Returns Duration of this range
+     *
+     * @return Duration
+     */
+    public function toDuration(): Duration
+    {
+        return Duration::ofDays($this->count());
```
BenMorel, line 238: Shouldn't this be `Duration::ofDays($this->count() - 1)`?

As it stands, `toPeriod()` returns a 0-day Period if the start date is equal to the end date (this is in line with its Java counterpart).

So wouldn't it be weird if it returned a 1-day Duration for, say, `2022-06-05/2022-06-05`?

solodkiy: > This object is iterable and countable: the iterator returns all the LocalDate objects contained
> in the range, while `count()` returns the total number of dates contained in the range.
 
> The end date, inclusive.

I think duration of one day range is One day, not zero)

solodkiy: >As it stands, toPeriod() returns a 0-day Period if the start date is equal to the end date (this is in line with its Java counterpart).

Quite weird. Is LocalDateRange in java inclusive too?

solodkiy:
I think about time zones and Local Range duration. In some time zones with Daylight Saving Time duration between day start and day end is not so obvious. 

For example Europe/London: 
> Sunday, 27 March 2022, 01:00:00 clocks were turned forward 1 hour to
> Sunday, 27 March 2022, 02:00:00 local daylight time instead.

So Duration of range 27 March 2022 to 28 March 2022 in London is 47h.
Should we allow pass TimeZone param to `toDuration()` method?

src/LocalDateRange.php
BenMorel: I think it makes sense in terms of Period: there is `1 day` between `2022-06-06` an `2022-06-07`.
In terms of Duration, maybe the approach is different. But this can be confusing.

solodkiy:
>What's your use case for this method exactly?

Actually I don't remember it anymore :)
I will return when I do.

PR #49, opened 2022-04-27
Make TimeZoneRegion::utc() return a TimeZoneRegion not an TimeZoneOffset

tigitz:
I think it's reasonable from a DX perspective to expect `TimeZoneRegion::utc()` to return a `TimeZoneRegion` instead of the offset.

### DIFF
--- a/src/Instant.php
+++ b/src/Instant.php
@@ -363,7 +363,7 @@ final class Instant implements JsonSerializable, Stringable
      */
     public function toISOString(): string
     {
-        return (string) ZonedDateTime::ofInstant($this, TimeZone::utc());
+        return (string) ZonedDateTime::ofInstant($this, TimeZoneOffset::utc());
     }
 
     /**
--- a/src/LocalDateTime.php
+++ b/src/LocalDateTime.php
@@ -715,7 +715,7 @@ final class LocalDateTime implements JsonSerializable, Stringable
      */
     public function toNativeDateTime(): DateTime
     {
-        return $this->atTimeZone(TimeZone::utc())->toNativeDateTime();
+        return $this->atTimeZone(TimeZoneOffset::utc())->toNativeDateTime();
     }
 
     /**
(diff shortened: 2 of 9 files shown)


tigitz:
@BenMorel If you could take a quick look about this one, it's pretty straight forward. It's slightly related to https://github.com/brick/date-time/issues/55

jiripudil:
Or perhaps there shouldn't be `TimeZone::utc()` at all, only leaving `TimeZoneOffset::utc()`? That's what Java does and it makes it super clear that it's an offset-based TZ.

I appreciate the shorthand form of `TimeZone::utc()` and I must admit we're using it a lot in our code, so both this PR and my proposal would be a breaking change, but if it would help clear some confusion, we could live with it; it's an easy change, after all.

BenMorel:
Good point about `TimeZoneRegion::utc()` returning an offset, we definitely need to fix this in one way or another.


I don't think keeping an abstract static `TimeZone::utc()` makes sense, it can probably be removed (in 0.5.0). 
That leaves us with two methods to represent a UTC time-zone: `TimeZoneOffset::utc()` and `TimeZoneRegion::utc()`. Isn't this confusing? What should we typically use here?

Indeed, Java has a `UTC` constant on [ZoneOffset](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html#UTC), but not [ZoneId](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html).

solodkiy:
Definitely don't want to see to different utc factory. Move `utc()` to `TimeZoneOffset` looks reasonable. 
Do we have any real use case for `TimeZoneRegion::utc()`?

BenMorel:
@solodkiy I don't think so. You can still construct it manually with `TimeZoneRegion::of('etc/UTC')` anyway if required.

BenMorel:
Moving forward, I don't think it hurts much to keep `TimeZone::utc()` for now, that delegates to `TimeZoneOffset::utc()` as it currently does.

@tigitz Can you please just add the new method in `TimeZoneRegion`, but not replace calls to `TimeZone::utc()` with `TimeZoneOffset::utc()`?

PR #53, opened 2022-06-08
Introduce coding style tool

tigitz:
Hello @BenMorel !

As discussed together, here's the first step towards contributing more and more.

Styling is always subject to preferences and I'm suggesting the ones I'm using in my own libs here. So feel free to make the final decision on each, whether you want to keep them or not.

I've chosen to `allow-risky` because IMO tests should cover any unexpected change of logic while using a formatting tool.

To ease the review, here's the list of the most controversial rules applied here:

- ordered_class_elements
- return_type_declaration (no space before)
- phpdoc_order (throws before return)
- native_function_invocation

You can find them here: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/index.rst
 
I've also split the Psalm and Unit tests into 2 workflows to parallelized the checks and halves execution time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,10 @@
 /vendor
 /composer.lock
 /.phpunit.result.cache
+
+# https://stackoverflow.com/a/16318111
+/tools/*
+!/tools/ecs
+/tools/ecs/*
+!/tools/ecs/composer.json
+!/tools/ecs/ecs.php
--- a/src/Clock/SystemClock.php
+++ b/src/Clock/SystemClock.php
@@ -14,11 +14,11 @@ use Brick\DateTime\Instant;
  */
 final class SystemClock implements Clock
 {
-    public function getTime() : Instant
+    public function getTime(): Instant
     {
         [$fraction, $epochSecond] = \explode(' ', microtime());
 
-        $epochSecond    = (int) $epochSecond;
+        $epochSecond = (int) $epochSecond;
         $nanoAdjustment = 10 * (int) \substr($fraction, 2, 8);
 
         return Instant::of($epochSecond, $nanoAdjustment);
--- a/src/DayOfWeek.php
+++ b/src/DayOfWeek.php
@@ -4,20 +4,22 @@ declare(strict_types=1);
 
 namespace Brick\DateTime;
 
+use JsonSerializable;
+
 /**
  * A day-of-week, such as Tuesday.
  *
  * This class is immutable.
  */
-final class DayOfWeek implements \JsonSerializable
+final class DayOfWeek implements JsonSerializable
 {
-    public const MONDAY    = 1;
-    public const TUESDAY   = 2;
+    public const MONDAY = 1;
+    public const TUESDAY = 2;
     public const WEDNESDAY = 3;
-    public const THURSDAY  = 4;
-    public const FRIDAY    = 5;
-    public const SATURDAY  = 6;
-    public const SUNDAY    = 7;
+    public const THURSDAY = 4;
+    public const FRIDAY = 5;
+    public const SATURDAY = 6;
+    public const SUNDAY = 7;
 
     /**
      * The ISO-8601 value for the day of the week, from 1 (Monday) to 7 (Sunday).
@@ -35,23 +37,6 @@ final class DayOfWeek implements \JsonSerializable
     }
 
     /**
-     * Returns a cached DayOfWeek instance.
-     *
-     * @param int $value The day-of-week value, validated from 1 to 7.
-     */
-    private static function get(int $value) : DayOfWeek
-    {
-        /** @var array<int, DayOfWeek> $values */
-        static $values = [];
-
-        if (! isset($values[$value])) {
-            $values[$value] = new DayOfWeek($value);
-        }
-
-        return $values[$value];
-    }
-
-    /**
      * Returns an instance of DayOfWeek for the given day-of-week value.
      *
      * @param int $dayOfWeek The day-of-week value, from 1 (Monday) to 7 (Sunday).
@@ -60,7 +45,7 @@ final class DayOfWeek implements \JsonSerializable
      *
      * @throws DateTimeException If the day-of-week is not valid.
      */
-    public static function of(int $dayOfWeek) : DayOfWeek
+    public static function of(int $dayOfWeek): DayOfWeek
     {
         Field\DayOfWeek::check($dayOfWeek);
 
@@ -72,7 +57,7 @@ final class DayOfWeek implements \JsonSerializable
      *
      * If no clock is provided, the system clock is used.
      */
-    public static function now(TimeZone $timeZone, ?Clock $clock = null) : DayOfWeek
+    public static function now(TimeZone $timeZone, ?Clock $clock = null): DayOfWeek
     {
         return LocalDate::now($timeZone, $clock)->getDayOfWeek();
     }
@@ -84,7 +69,7 @@ final class DayOfWeek implements \JsonSerializable
      *
      * @return DayOfWeek[]
      */
-    public static function all(?DayOfWeek $first = null) : array
+    public static function all(?DayOfWeek $first = null): array
     {
         $days = [];
         $first = $first ?: DayOfWeek::get(DayOfWeek::MONDAY);
@@ -93,8 +78,7 @@ final class DayOfWeek implements \JsonSerializable
         do {
             $days[] = $current;
             $current = $current->plus(1);
-        }
-        while (! $current->isEqualTo($first));
+        } while (! $current->isEqualTo($first));
 
         return $days;
     }
@@ -102,7 +86,7 @@ final class DayOfWeek implements \JsonSerializable
     /**
      * Returns a day-of-week instance for Monday.
      */
-    public static function monday() : DayOfWeek
+    public static function monday(): DayOfWeek
     {
         return DayOfWeek::get(DayOfWeek::MONDAY);
     }
@@ -110,7 +94,7 @@ final class DayOfWeek implements \JsonSerializable
     /**
      * Returns a day-of-week instance for Tuesday.
      */
-    public static function tuesday() : DayOfWeek
+    public static function tuesday(): DayOfWeek
     {
         return DayOfWeek::get(DayOfWeek::TUESDAY);
     }
@@ -118,7 +102,7 @@ final class DayOfWeek implements \JsonSerializable
     /**
      * Returns a day-of-week instance for Wednesday.
      */
-    public static function wednesday() : DayOfWeek
+    public static function wednesday(): DayOfWeek
     {
         return DayOfWeek::get(DayOfWeek::WEDNESDAY);
     }
@@ -126,7 +110,7 @@ final class DayOfWeek implements \JsonSerializable
     /**
      * Returns a day-of-week instance for Thursday.
      */
-    public static function thursday() : DayOfWeek
+    public static function thursday(): DayOfWeek
     {
         return DayOfWeek::get(DayOfWeek::THURSDAY);
     }
@@ -134,7 +118,7 @@ final class DayOfWeek implements \JsonSerializable
     /**
      * Returns a day-of-week instance for Friday.
      */
-    public static function friday() : DayOfWeek
+    public static function friday(): DayOfWeek
     {
         return DayOfWeek::get(DayOfWeek::FRIDAY);
     }
@@ -142,7 +126,7 @@ final class DayOfWeek implements \JsonSerializable
     /**
      * Returns a day-of-week instance for Saturday.
      */
-    public static function saturday() : DayOfWeek
+    public static function saturday(): DayOfWeek
     {
         return DayOfWeek::get(DayOfWeek::SATURDAY);
     }
@@ -150,7 +134,7 @@ final class DayOfWeek implements \JsonSerializable
     /**
      * Returns a day-of-week instance for Sunday.
      */
-    public static function sunday() : DayOfWeek
+    public static function sunday(): DayOfWeek
     {
         return DayOfWeek::get(DayOfWeek::SUNDAY);
     }
@@ -160,7 +144,7 @@ final class DayOfWeek implements \JsonSerializable
      *
      * @return int The day-of-week value, from 1 (Monday) to 7 (Sunday).
      */
-    public function getValue() : int
+    public function getValue(): int
     {
         return $this->value;
     }
@@ -172,7 +156,7 @@ final class DayOfWeek implements \JsonSerializable
      *
      * @return bool True if this day-of-week is equal to the given value, false otherwise.
      */
-    public function is(int $dayOfWeek) : bool
+    public function is(int $dayOfWeek): bool
     {
         return $this->value === $dayOfWeek;
     }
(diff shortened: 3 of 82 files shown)


src/DayOfWeek.php
```diff
@@ -93,64 +92,63 @@ public static function all(?DayOfWeek $first = null) : array
         do {
             $days[] = $current;
             $current = $current->plus(1);
-        }
-        while (! $current->isEqualTo($first));
+        } while (!$current->isEqualTo($first));
```
BenMorel, line 95: If possible, I'd like to keep the space here.

src/DayOfWeek.php
```diff
@@ -184,64 +182,65 @@ public function is(int $dayOfWeek) : bool
...
     }
 
     /**
      * Returns the DayOfWeek that is the specified number of days before this one.
      */
-    public function minus(int $days) : DayOfWeek
+    public function minus(int $days): DayOfWeek
     {
-        return $this->plus(- $days);
+        return $this->plus(-$days);
```
BenMorel, line 219: Doubt on this one. Does that mean that we write `-$b`, but `$a - $b`?

src/Month.php
```diff
@@ -105,25 +107,25 @@ public function getValue() : int
...
     }
 
     /**
      * Returns whether this Month equals another Month.
      */
-    public function isEqualTo(Month $that) : bool
+    public function isEqualTo(Month $that): bool
     {
-        return ($this->month === $that->month);
+        return $this->month === $that->month;
```
BenMorel, line 120: OK with that.

BenMorel:
Thanks for the PR, @tigitz! I haven't finished the review yet, but I think you've got 90% of the relevant points already. Enough to keep you busy a bit :)

src/DayOfWeek.php
```diff
@@ -35,32 +35,31 @@ private function __construct(int $value)
...
     /**
      * Returns an instance of DayOfWeek for the given day-of-week value.
      *
      * @param int $dayOfWeek The day-of-week value, from 1 (Monday) to 7 (Sunday).
      *
-     * @return DayOfWeek The DayOfWeek instance.
-     *
      * @throws DateTimeException If the day-of-week is not valid.
+     *
+     * @return DayOfWeek The DayOfWeek instance.
```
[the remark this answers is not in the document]
tigitz: Unfortunately not: https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/phpdoc/phpdoc_order.rst

I'll remove the rule.

src/DayOfWeek.php
tigitz: Yes, here are the two rules in play here:
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/operator/unary_operator_spaces.rst
https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/operator/binary_operator_spaces.rst

.php-cs-fixer.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
+$finder = (new PhpCsFixer\Finder())
+    ->in([__DIR__.'/src', __DIR__.'/tests'])
```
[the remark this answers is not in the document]
tigitz: Sure

src/DayOfWeek.php
tigitz: OK, added `not_operator_with_successor_space` rule

tigitz:
@BenMorel Updated !

I took the opportunity to introduced two new rules:

[self_accessor](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/class_notation/self_accessor.rst)
[php_unit_strict](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/php_unit/php_unit_strict.rst)

src/Clock/SystemClock.php
```diff
@@ -14,12 +14,12 @@
...
-    public function getTime() : Instant
+    public function getTime(): Instant
     {
-        [$fraction, $epochSecond] = \explode(' ', microtime());
+        [$fraction, $epochSecond] = explode(' ', microtime());
 
-        $epochSecond    = (int) $epochSecond;
-        $nanoAdjustment = 10 * (int) \substr($fraction, 2, 8);
+        $epochSecond = (int) $epochSecond;
+        $nanoAdjustment = 10 * (int) substr($fraction, 2, 8);
```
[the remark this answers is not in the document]
BenMorel: The impact on readability is only negative if you use a backslash before the function name. If you `use function` at the top of the file, it's close to zero.

Doctrine [does this](https://github.com/doctrine/orm/blob/3dc5581294ccc05e0150c8b6708ccefbea88849e/lib/Doctrine/ORM/EntityManager.php#L37-L48), but they use phpcs :/

BenMorel:
Just reviewed the `self_accessor` rule, and I think it actually hurts readability!

```php
class DayOfWeek
{
    public function isEqualTo(self $that): bool {
```

vs:

```php
class DayOfWeek
{
    public function isEqualTo(DayOfWeek $that): bool {
```

Mixing both styles is not good though, is there a rule that actually does the opposite?

tests/InstantTest.php
```diff
@@ -448,60 +448,60 @@ public function providerCompareTo() : array
             [-1,  1,  1, -1, -1],
             [-1,  1,  1,  0, -1],
             [-1,  1,  1,  1, -1],
-            [ 0, -1, -1, -1,  1],
```
BenMorel, line 451: We'll have fun discussing this rule in brick/math :)

https://github.com/brick/math/blob/ec750c1b5bf4158b3823309bbe1887b40e837c8d/tests/BigIntegerTest.php#L2329-L2389

(spoiler: once settled on a coding standard, I'd like to apply it to all projects)

Not sure how we can reconcile readability & coding standards in large tests like these. Thoughts?

src/Parser/IsoParsers.php
```diff
@@ -116,17 +116,17 @@ public static function localDateRange() : PatternParser
...
+            ->append(self::localDate())
+            ->endGroup()
             ->appendOr()
-                ->startGroup()
-                    ->startOptional()
-                        ->appendCapturePattern(MonthOfYear::PATTERN, MonthOfYear::NAME)
-                        ->appendLiteral('-')
-                    ->endOptional()
-                    ->appendCapturePattern(DayOfMonth::PATTERN, DayOfMonth::NAME)
-                ->endGroup()
```
BenMorel, line 129: Same as in tests: I do value coding standards, but I'd really miss these indents for readability. Ideas?

tigitz:
@BenMorel Rebased and introduce a different tool (https://github.com/symplify/easy-coding-standard) to be able to use both phpcsfixer rules and phpcs sniffs together as it seems to be required here.

I've removed the `self_accessor` rule.

Thanks to PHPCS sniffs availability it can now put `@throws` after `@return` in doc blocks like initially requested here: https://github.com/brick/date-time/pull/53#discussion_r892892182

It can also ignores rules and sniffs on specific files which is something I used to provide a solution to your request here: https://github.com/brick/date-time/pull/53#discussion_r892892182

I've introduced the same sniff doctrine is using to provide a solution to your request here: https://github.com/brick/date-time/pull/53#discussion_r895071370

Let me know if there's any blockers left to tackle. The sooner it's merged the less I'll have to go through the rebasing / conflict management process :sweat_smile:

src/Parser/IsoParsers.php
tigitz: Fix by https://github.com/brick/date-time/pull/53#issuecomment-1153311045

BenMorel:
Thanks, @tigitz! A couple questions:

- what's the advantage of installing tools in a separate directory / composer.json? vs. installing using `require-dev`? One drawback I can see is that you have to perform 2 `composer install`s when contributing to the project. And what about Psalm?
- can ecs ( / phpcs) fix automatically the codebase, like php-cs-fixer does?

src/LocalDateTime.php
```diff
@@ -734,20 +695,60 @@ public function toDateTime() : \DateTime
...
+        $totDays += Math::floorDiv($totSeconds, LocalTime::SECONDS_PER_DAY);
+        $newSoD = Math::floorMod($totSeconds, LocalTime::SECONDS_PER_DAY);
+
+        $newDate = $this->date->plusDays($totDays);
+        $newTime = ($newSoD === $curSoD ? $this->time : LocalTime::ofSecondOfDay($newSoD, $newNano));
+
+        return new LocalDateTime($newDate, $newTime);
+    }
+
+    public function __toString(): string
```
BenMorel, line 751: Can we have the magic methods at the end of the public methods, instead of after private methods? I don't like it this way.

If not possible, second best choice: back to after `__construct()`.

src/TimeZone.php
```diff
@@ -53,45 +54,46 @@ abstract public function getId() : string;
...
-    public static function fromNativeDateTimeZone(\DateTimeZone $dateTimeZone) : TimeZone
+    public static function fromNativeDateTimeZone(DateTimeZone $dateTimeZone): TimeZone
     {
         return TimeZone::parse($dateTimeZone->getName());
     }
 
     /**
      * Returns an equivalent native `DateTimeZone` object for this TimeZone.
+     *
      * @deprecated please use toNativeDateTimeZone instead
```
BenMorel, line 80: Is there a way to force each `@param`, `@return`, `@throws`, `@deprecated` sentence to end with a punctuation mark?

tests/DurationTest.php
```diff
@@ -861,8 +863,8 @@ public function providerMultipliedBy() : array
             [3, 0, 3, 9, 0],
             [3, 1000, 3, 9, 3000],
             [3, 999999999, 3, 11, 999999997],
-            [1, 0,  \PHP_INT_MAX, \PHP_INT_MAX, 0],
-            [1, 0, \PHP_INT_MIN, \PHP_INT_MIN, 0],
+            [1, 0,  PHP_INT_MAX, PHP_INT_MAX, 0],
```
BenMorel, line 866: Weird that the double-space has not been caught here?

BenMorel (changes requested):
Thanks a lot for your work, @tigitz! The codebase looks in much better shape now.

I've completed the review, let's fix/discuss the last points and it'll be good for merge. If you can complete soon enough, I can stop commits/merges in the meantime to avoid conflicts.

BenMorel:
By the way, I understand, and can learn to like, the `tools/` approach. 
> Here's a more in-depth discussion about the problem: https://github.com/composer/composer/issues/9636

Thanks for the pointer, I actually tried `bamarni/composer-bin-plugin` but apart from the symlink in the main `vendor/bin`directory (which the current maintainer said he doesn't use himself), I wonder what it brings vs your "manual" approach? Maybe the convenience of `composer bin all install`/`update`?

> If you validate the approach, I can migrate psalm the same way. In this PR's scope or in an other. Up to you.

Let's do this in a separate PR, along with some instructions in the README on how to setup the whole thing, for new contributors?

ecs.php
```diff
@@ -0,0 +1,216 @@
+<?php
```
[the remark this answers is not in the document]
tigitz: Done

.gitignore
```diff
@@ -1,3 +1,9 @@
 /vendor
 /composer.lock
 /.phpunit.result.cache
+
+# https://stackoverflow.com/a/16318111
+/tools/*
+!/tools/ecs
+/tools/ecs/*
+!/tools/ecs/composer.json
```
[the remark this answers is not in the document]
tigitz: Unfortunately not, I've put a stackoverflow link above to "document" why it's required: https://stackoverflow.com/a/16318111

src/Clock/SystemClock.php
```diff
@@ -6,6 +6,8 @@
 
 use Brick\DateTime\Clock;
 use Brick\DateTime\Instant;
+use function explode;
```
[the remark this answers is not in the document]
tigitz: Yep, I've added it

src/LocalDateTime.php
tigitz: Yep, I've added it. However I can only order it like this ` 'construct', 'phpunit', 'method_public', 'magic', 'method_protected', 'method_private'` which means that if magic methods are protected or private they'll be after public methods anyway. Since it's quite rare I guess it's ok.

src/TimeZone.php
tigitz: Forcing no. However I removed https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/phpdoc/phpdoc_annotation_without_dot.rst to allow it.

tigitz: I eventually found a rule that forces the dot on `@param` and `@throws` annotations. It also cannot be automatically fixed. That's better than nothing I guess.

tigitz:
@BenMorel Answered, updated and rebased.

> I wonder what it brings vs your "manual" approach? Maybe the convenience of composer bin all install/update?

From my limited knowledge on the tool I would say yes. I don't think it's worth adding another "tool" to manage other tools while composer can already do the trick.

> Let's do this in a separate PR, along with some instructions in the README on how to setup the whole thing, for new contributors?

I've added a `Contributing` section to the README.md

tests/DurationTest.php
BenMorel: I'm interested in having this rule *most of the time*, but to be able to exclude it when done on purpose to improve readability, like in some of the tests. As you've done for other rules?

BenMorel:
@tigitz Thank you, let's figure out the last two points and let's merge!

.gitignore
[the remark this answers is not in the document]
tigitz: I've tested both version too on my hand and the result is quite different:

image

image

I thought it could be a git version problem but upgrading from `2.25` to `2.36` didn't change the behavior.

What makes you think they work equally fine ?

tests/DurationTest.php
tigitz: @BenMorel Understood, however in this specific case the `DurationTest` file has column-aligned formatted arrays which shouldn't be touched so excluding the file is necessary and that will make this particular issue not detected and therefore not fixed. 

The only way to exclude the check for specific arrays in the same file would be to use a more fine grained approach using php comments

```php
// phpcs:ignore Squiz.Arrays.ArrayDeclaration.SingleLineNotAllowed
```

But I'm not sure the maintenance burden of applying these comments for every arrays is worth it. WDYT ?

For now I've excluded the whole `tests` folder from this `DuplicateSpacesSniff`

.gitignore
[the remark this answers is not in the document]
BenMorel: Weird, I can't reproduce your issue here. When I remove these 2 lines, git still tracks `composer.json` and `ecs.php`.
I guess it doesn't hurt much to leave them, though.

tests/DurationTest.php
BenMorel: I guess it's good enough for now, then!

BenMorel:
Thanks a lot, @tigitz! If you're interested, we can think about how to deploy the same coding standards to other brick projects, without duplicating the config, composer.json etc. every time. Maybe a separate `brick/coding-standard` repo that we can use in each library repo?

tigitz:
@BenMorel happy to contribute 😊

Let's do this yeah. This what doctrine and sylius are doing. I'll take inspiration from them.

Ping me when repo is created, thanks.

BenMorel:
Here you are: https://github.com/brick/coding-standards :slightly_smiling_face:

Issue #47, opened 2022-04-10
ZonedDateTime::toDateTime and getDateTime conflict


solodkiy:
In ZonedDateTime class we have two methods which looks almost the same, but different in return types:

-  `public function toDateTime() : \DateTime`
-  `public function getDateTime() : LocalDateTime`

To avoid misunderstanding and confuse I propose rename `toDateTime` to `toNativeDateTime`.
This approach could be used in other conversions in and from native types. For example:

- `Period::fromDateInterval` -> `Period::fromNativeDateInterval`
- `Period::toDateInterval` -> `Period::toNativeDateInterval`
- `LocalDateRange::toDatePeriod` -> `LocalDateRange::toNativeDatePeriod`
- `TimeZone::fromDateTimeZone` -> `TimeZone::fromNativeDateTimeZone`
etc..

Another variant is use Php prefix (`toDateTime` -> `toPhpDateTime`).

tigitz:
IMO, the `Native` prefix makes sense, I'm :+1: on this it helps readability and reduce ambiguity. I've wondered myself while trying to autocomplete method calls if the suggested `DatePeriod` for example was a library class or the native one.

However I would keep the `getDateTime` as it is since it means it is `getting` the DateTime part of the `ZonedDateTime`. Using the `to` prefix would suggest it's transforming the object into another one which is not the case here.

@BenMorel wdyt ?

solodkiy:
Some of `get*`/`to*()` discussion was there: https://github.com/brick/date-time/pull/48

BenMorel:
LGTM as well with the `Native` prefix. @solodkiy You're welcome to open a PR to create the new `(to|from)Native*()` methods, and deprecate existing ones. We'll remove them in `0.5.0`.

The `getDateTime()` method is in line with the discussion we had in #48, so it looks good as it is!

PR #56, opened 2022-06-10
LocalDateRange::toNativeDatePeriod

solodkiy:
LocalDateRange::toNativeDatePeriod

https://github.com/brick/date-time/issues/47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,12 +12,13 @@
 
 💩 **Deprecations**
 
-The following methods have been deprecated in favour of new names:
+The following methods have been deprecated in favour of new names ([#47](https://github.com/brick/date-time/issues/47)):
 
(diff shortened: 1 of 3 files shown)


BenMorel:
LGTM, could you please update the CHANGELOG? And in #57 and #58 as well.

solodkiy:
@BenMorel I tried to use same emoji, but I cannot see it in macos. Maybe we should use old good plain text instead?
image

BenMorel:
@solodkiy It's a Unicode character that I thought would be quite universal: https://unicode-table.com/en/1F872/

Feel free to replace with another arrow that works for you: https://unicode-table.com/en/sets/arrow-symbols/, or just with `->` if you still have trouble (replace the others as well for consistency).

solodkiy:
Changed to [U+2192](https://unicode-table.com/en/2192/)

BenMorel:
Thanks :+1:

PR #60, opened 2022-06-19
Bring back support for sub-minute offsets in PHP >= 8.1.7

### LINKED ISSUE #44: `ZonedDateTime::of()` issue when year is less than 1892
Following code works well (no output):

```php
ZonedDateTime::of(
	LocalDateTime::of(2022,1,1),
	TimeZoneRegion::parse('Europe/Prague')
);
```

while this one throws `Brick\DateTime\DateTimeException: The time zone offset of 3464 seconds is not a multiple of 60`:

```php
ZonedDateTime::of(
	LocalDateTime::of(0,1,1),
	TimeZoneRegion::parse('Europe/Prague')
);
```

The only difference is the year passed into `LocalDateTime::of()`. If I try `LocalDateTime::min()` instead, it results in the exception as well.

I found out that the border where it stops working is 1891–1892 (which 130 years back from now). Is this some issue with `DateTime` and/or its `getOffset()` which is used inside of the `ZonedDateTime::of()`?

But it doesn't fail with UTC time zone actually..

jiripudil:
closes #44

### DIFF
--- a/tests/ZonedDateTimeTest.php
+++ b/tests/ZonedDateTimeTest.php
@@ -394,9 +394,9 @@ class ZonedDateTimeTest extends AbstractTestCase
         ZonedDateTime::parse($text);
     }
 
-    public function providerParseInvalidStringThrowsException(): array
+    public function providerParseInvalidStringThrowsException(): iterable
     {
-        return [
+        yield from [
             [''],
             ['2001'],
             ['2001-'],
@@ -418,6 +418,12 @@ class ZonedDateTimeTest extends AbstractTestCase
             [' 2001-02-03T01:02:03Z'],
             ['2001-02-03T01:02:03Z '],
         ];
+
+        if (PHP_VERSION_ID < 80107) {
+            yield from [
+                ['2001-02-03T01:02:03.456+12:34:56'],
+            ];
+        }
     }
 
     /**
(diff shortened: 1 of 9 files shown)


tests/ZonedDateTimeTest.php
```diff
@@ -408,7 +414,6 @@ public function providerParseInvalidStringThrowsException(): array
             ['2001-02-03T04:05:06[Europe/London]'],
             ['2001-02-03T04:05.789Z[Europe/London]'],
             ['2001-02-03T04:05:06Z[Europe/London'],
-            ['2001-02-03T01:02:03.456+12:34:56'],
```
BenMorel, line 411: Could you please keep this test for PHP < 8.1.7, as you've done for other tests?

BenMorel (changes requested):
Thanks a lot for the quality work, and sorry for the delay @jiripudil!
Just a couple nitpicks and I'll merge it :+1:

BenMorel:
By the way, regarding the workaround for `DateTimeZone::getName()` omitting seconds, I created an issue with PHP:

https://github.com/php/php-src/issues/11281

jiripudil:
Hi, thanks, I've fixed the nitpicks and rebased the branch.

Also it seems that the issue has been fixed already (https://github.com/php/php-src/commit/f9117eb82487a0566c8fee3fda798cc2eb969d4a), so I've updated the workaround to only apply in affected PHP versions.

BenMorel (approved):

BenMorel:
Thank you, @jiripudil !

PR #64, opened 2022-08-12
Implement contains, intersection and `of` methods for Interval

### LINKED ISSUE #63: `Interval` is so barebones compared to `LocalDateRange`
I suggest `Interval` should have the following methods, like in `LocalDateRange`:

- `contains(Instant): bool`
- `intersectsWith(Interval): bool`
- `getIntersectionWith(Interval): Interval`
- `of(Instant, Instant): self` - for consistency with other classes

and nice to haves (also contained in the `LocalDateRange`):
- `isEqualTo(Interval): bool`
- `toNativeDatePeriod(): \DatePeriod`

someniatko:
closes #63

### DIFF
--- a/src/Interval.php
+++ b/src/Interval.php
@@ -41,6 +41,17 @@ final class Interval implements JsonSerializable
     }
 
     /**
+     * @param Instant $startInclusive The start instant, inclusive.
+     * @param Instant $endExclusive   The end instant, exclusive.
+     *
+     * @throws DateTimeException If the end instant is before the start instant.
+     */
+    public static function of(Instant $startInclusive, Instant $endExclusive): Interval
+    {
+        return new Interval($startInclusive, $endExclusive);
+    }
+
+    /**
      * Returns the start instant, inclusive, of this Interval.
      */
     public function getStart(): Instant
@@ -82,6 +93,44 @@ final class Interval implements JsonSerializable
     public function getDuration(): Duration
     {
         return Duration::between($this->start, $this->end);
+    }
+
+    /**
+     * Returns whether this Interval contains the given Instant.
+     */
+    public function contains(Instant $instant): bool
+    {
+        return $instant->isAfterOrEqualTo($this->start)
+            && $instant->isBefore($this->end);
+    }
+
+    /**
+     * Returns whether this Interval intersects with the given one.
+     */
+    public function intersectsWith(Interval $that): bool
+    {
+        [$prev, $next] = $this->start->isBefore($that->start)
+            ? [$this, $that]
+            : [$that, $this];
+
+        return $next->start->isBefore($prev->end);
+    }
+
+    /**
+     * Returns an Interval which is an intersection of this one with the given one.
+     *
+     * @throws DateTimeException If the Intervals do not intersect.
+     */
+    public function getIntersectionWith(Interval $that): Interval
+    {
+        if (! $this->intersectsWith($that)) {
+            throw new DateTimeException('Intervals "' . $this . '" and "' . $that . '" do not intersect.');
+        }
+
+        $latestStart = $this->start->isAfter($that->start) ? $this->start : $that->start;
+        $earliestEnd = $this->end->isBefore($that->end) ? $this->end : $that->end;
+
+        return new Interval($latestStart, $earliestEnd);
     }
 
     /**
--- a/src/Interval.php
+++ b/src/Interval.php
@@ -25,6 +25,8 @@ final class Interval implements JsonSerializable
     private Instant $end;
 
     /**
+     * @deprecated Use {@see Interval::of()} instead.
+     *
      * @param Instant $startInclusive The start instant, inclusive.
      * @param Instant $endExclusive   The end instant, exclusive.
      *
@@ -74,7 +76,7 @@ final class Interval implements JsonSerializable
      */
     public function withStart(Instant $start): Interval
     {
-        return new Interval($start, $this->end);
+        return Interval::of($start, $this->end);
     }
 
     /**
@@ -84,7 +86,7 @@ final class Interval implements JsonSerializable
      */
     public function withEnd(Instant $end): Interval
     {
-        return new Interval($this->start, $end);
+        return Interval::of($this->start, $end);
     }
 
     /**
@@ -130,7 +132,7 @@ final class Interval implements JsonSerializable
         $latestStart = $this->start->isAfter($that->start) ? $this->start : $that->start;
         $earliestEnd = $this->end->isBefore($that->end) ? $this->end : $that->end;
 
-        return new Interval($latestStart, $earliestEnd);
+        return Interval::of($latestStart, $earliestEnd);
     }
 
     public function isEqualTo(Interval $that): bool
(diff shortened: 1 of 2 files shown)


someniatko:
Do you think it makes sense to also implement `isEqualTo()` and `toNativeDatePeriod()`?

someniatko:
@BenMorel ping 🤸

tests/IntervalTest.php
```diff
@@ -95,6 +106,135 @@ public function testGetDuration(): void
...
+                5000, 6000,
+            ],
+        ];
+    }
+
+    public function testGetIntersectionWithInvalidParams(): void
+    {
+        $interval1 = new Interval(Instant::of(100000), Instant::of(200000));
+        $interval2 = new Interval(Instant::of(300000), Instant::of(400000));
+        $this->expectException(DateTimeException::class);
```
BenMorel, line 234: Can you add an expectation for the exception message?

src/Interval.php
```diff
@@ -40,6 +40,17 @@ public function __construct(Instant $startInclusive, Instant $endExclusive)
         $this->end = $endExclusive;
     }
 
+    /**
+     * @param Instant $startInclusive The start instant, inclusive.
+     * @param Instant $endExclusive   The end instant, exclusive.
+     *
+     * @throws DateTimeException If the end instant is before the start instant.
+     */
+    public static function of(Instant $startInclusive, Instant $endExclusive): Interval
```
BenMorel, line 49: Maybe we should deprecate the public constructor in favour of this factory method, to keep it consistent with other classes?

tests/IntervalTest.php
```diff
@@ -95,6 +106,135 @@ public function testGetDuration(): void
         $this->assertDurationIs(1, 999444556, $duration);
     }
 
+    /** @dataProvider providerContains */
+    public function testContains(int $start, int $end, int $now, bool $expected, string $errorMessage): void
+    {
+        $interval = new Interval(Instant::of($start), Instant::of($end));
+
+        $this->assertEquals($expected, $interval->contains(Instant::of($now)), $errorMessage);
```
BenMorel, line 114: `assertSame()` please

tests/IntervalTest.php
```diff
@@ -95,6 +106,135 @@ public function testGetDuration(): void
...
+            ],
+        ];
+    }
+
+    /** @dataProvider providerIntersectsWith */
+    public function testIntersectsWith(int $start1, int $end1, int $start2, int $end2, bool $expected): void
+    {
+        $interval1 = new Interval(Instant::of($start1), Instant::of($end1));
+        $interval2 = new Interval(Instant::of($start2), Instant::of($end2));
+        $this->assertEquals($expected, $interval1->intersectsWith($interval2));
```
BenMorel, line 149: `assertSame()`

tests/IntervalTest.php
```diff
@@ -95,6 +106,135 @@ public function testGetDuration(): void
...
+        int $start2,
+        int $end2,
+        int $expectedStart,
+        int $expectedEnd
+    ): void {
+        $interval1 = new Interval(Instant::of($start1), Instant::of($end1));
+        $interval2 = new Interval(Instant::of($start2), Instant::of($end2));
+        $expected = new Interval(Instant::of($expectedStart), Instant::of($expectedEnd));
+
+        $this->assertEquals($expected, $interval1->getIntersectionWith($interval2));
```
BenMorel, line 196: I'm not keen on using `assertEquals()` for objects. Suggestions:

- either `assertSame()` their `(string)` representation
- or implement `Interval::isEqualTo(Interval $that)`, and use it to `assertTrue()` here

BenMorel (changes requested):
A couple changes, otherwise looks good to me :+1:

BenMorel (changes requested):
A couple changes, otherwise looks good to me :+1:

BenMorel:
> Do you think it makes sense to also implement isEqualTo() and toNativeDatePeriod()?

- `isEqualTo()`: yes, and you may use it in your tests as per my [comment](https://github.com/brick/date-time/pull/64#discussion_r1197047746)
- `toNativeDatePeriod()`: it looks like `DatePeriod` is about recurring periods, so I don't think they represent the same concept. I'd pass on this one.

src/Interval.php
someniatko: Yeah, it makes sense

tests/IntervalTest.php
someniatko: Yeah, until PHP implements proper equality override for objects, which would be very useful for value objects, it's indeed safer to use other means, although I thought it is okay enough here. By the way I dunno why it still doesn't happen, I've already seen a couple of RFCs on the topic, but haven't dived into the exact rejection reasons.

someniatko:
Okay, I will implement `isEqualTo()` as well. And yeah, you are right about `toNativeDatePeriod()` (should be called `toNativeDateInterval()`) - it indeed does not belong here, it's more fit for the `Duration` class, like in the `Period`.

someniatko:
Done! Code coverage has been decreased for the classes this PR is not responsible for, maybe something has changed in the method of the coverage calculation.

tests/IntervalTest.php
someniatko: added

tests/IntervalTest.php
```diff
@@ -23,15 +23,26 @@ public function testEndInstantIsNotBeforeStartInstant(): void
...
-        new Interval($end, $start);
+        Interval::of($end, $start);
     }
 
     public function testGetStartEnd(): void
     {
         $start = Instant::of(2000000000, 987654321);
         $end = Instant::of(2000000009, 123456789);
 
-        $interval = new Interval($start, $end);
```
BenMorel, line 34: You should keep `new` here, otherwise the test is the same as `testGetStartEndUsingOf()`!

src/Interval.php
```diff
@@ -25,6 +25,8 @@ final class Interval implements JsonSerializable
     private Instant $end;
 
     /**
+     * @deprecated Use {@see Interval::of()} instead, for the consistency with other classes in this library.
```
BenMorel, line 28: Just:

```
@deprecated Use {@see Interval::of()} instead.
```

please!

BenMorel (changes requested):

tests/IntervalTest.php
someniatko: Ah yes, haven't noticed it

someniatko: Since the constructor is deprecated, I think we can remove `testGetStartEndUsingOf()` test. I can add a single test which specifically tests that the public constructor still works though.

src/Interval.php
someniatko: okay, let's not disclose our intents to the potential enemy :D

tests/IntervalTest.php
BenMorel: > I can add a single test which specifically tests that the public constructor still works though.

That's this test :slightly_smiling_face: But you can rename it to make it more obvious.

BenMorel (approved):

BenMorel:
Thank you! Released as [0.4.2](https://github.com/brick/date-time/releases/tag/0.4.2) :tada:

PR #75, opened 2023-08-16
Updated src : created TimeStamp class

pj8912:
With TimeStamp class :
- get current timestamp
- create new timestamp


BenMorel:
Hi, I don't think that this class belongs to this library.

We already have `Instant` that represents an immutable UNIX timestamp with nanosecond precision!

PR #77, opened 2023-09-14
[performance] Use static variables inside every singleton constructors.

gnutix:
Hello folks,

Here's another performance optimization when static constructors are called a lot. It was already done for some methods in the library, so I kept the same code style and applied it to all static constructors that take no arguments ("constants", in a way).

Let me know if you think some of these do not deserve it. I applied it everywhere without much thinking about it, because I don't think there's any downside of having that static variable, and I definitely know that some of these methods show up a lot in our benchmarks, like `TimeZone(Offset)::utc()` and `Duration::zero()`, `min()/max()`, etc, and then it's definitely beneficial.

### DIFF
--- a/src/Instant.php
+++ b/src/Instant.php
@@ -77,7 +77,14 @@ final class Instant implements JsonSerializable
 
     public static function epoch(): Instant
     {
-        return new Instant(0, 0);
+        /** @var Instant|null $epoch */
+        static $epoch;
+
+        if ($epoch) {
+            return $epoch;
+        }
+
+        return $epoch = new Instant(0, 0);
     }
 
     public static function now(?Clock $clock = null): Instant
@@ -96,7 +103,14 @@ final class Instant implements JsonSerializable
      */
     public static function min(): Instant
     {
-        return new Instant(PHP_INT_MIN, 0);
+        /** @var Instant|null $min */
+        static $min;
+
+        if ($min) {
+            return $min;
+        }
+
+        return $min = new Instant(PHP_INT_MIN, 0);
     }
 
     /**
@@ -106,7 +120,14 @@ final class Instant implements JsonSerializable
      */
     public static function max(): Instant
     {
-        return new Instant(PHP_INT_MAX, 999_999_999);
+        /** @var Instant|null $max */
+        static $max;
+
+        if ($max) {
+            return $max;
+        }
+
+        return $max = new Instant(PHP_INT_MAX, 999_999_999);
     }
 
     public function plus(Duration $duration): Instant
--- a/src/Duration.php
+++ b/src/Duration.php
@@ -54,7 +54,14 @@ final class Duration implements JsonSerializable
      */
     public static function zero(): Duration
     {
-        return new Duration(0);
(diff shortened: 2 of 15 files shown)


gnutix:
Tests added, reverted `DayOfWeek` and `Month` as suggested.

PS: i find it unexpected that ECS is configured with `PhpCsFixer\Fixer\PhpUnit\PhpUnitTestCaseStaticMethodCallsFixer` so that it forces using `$this->` to call methods that are actually static (like `assertSame`).

gnutix:
Looking at some Blackfire inspections, I'm thinking `__toString()` method might benefit from the same static variable trick. Would you accept a similar PR for these ?

BenMorel:
> PS: i find it unexpected that ECS is configured with `PhpCsFixer\Fixer\PhpUnit\PhpUnitTestCaseStaticMethodCallsFixer` so that it forces using `$this->` to call methods that are actually static (like `assertSame`). :man_shrugging:

I tend to agree. @tigitz Do you want to change this?

tigitz:
Yeah it's a matter of preference as explained here: https://docs.phpunit.de/en/10.3/assertions.html

The idea I guess from this rule would be to at least force consistency throughout the codebase.

But if you don't mind having both usage then I can remove it. Let me know. 

@BenMorel Taking the opportunity to friendly remind you about the dedicated PR about CI and codestyle that is still pending your review 😉

BenMorel:
Thank you, @gnutix!

@tigitz:

> The idea I guess from this rule would be to at least force consistency throughout the codebase.

Definitely! I was thinking about switching to `self`, not allowing both. After double checking, this is actually the style I already use in brick/math and brick/money at least.

> Taking the opportunity to friendly remind you about the dedicated PR about CI and codestyle that is still pending your review 😉

I know, I'm sorry, I'm still a bit unsure about whether this is the way I want things, so I postponed indefinitely 😕 But the lack of coding standards on the other libraries really bothers me, so I'll have to take a decision at some point.

In the meantime, if you want to change the PHPUnit code style, a PR is welcome :slightly_smiling_face:

src/Instant.php
```diff
@@ -77,7 +77,14 @@ public static function of(int $epochSecond, int $nanoAdjustment = 0): Instant
...
     {
-        return new Instant(0, 0);
+        /** @var Instant|null $epoch */
+        static $epoch;
+
+        if ($epoch) {
+            return $epoch;
+        }
+
+        return $epoch = new Instant(0, 0);
```
solodkiy, line 87:
```php
        /** @var Instant|null $epoch */
        static $epoch;

        if (!isset($epoch)) {
            $epoch = new Instant(0, 0);
        }

        return $epoch;
```
I think this variant is simpler, more cache-way

gnutix: Agree. That's how I spontaneously wrote the PR, then I changed it to match the current code style..

BenMorel: I don't mind if you want to change this.
If you do, please use `if ($epoch === null)` instead of `isset()`, which wouldn't alert on undefined variable!

PR #79, opened 2023-09-15
Optimize performance of LocalDate::plusDays(1) calls.

### LINKED ISSUE #78: [performance] Optimizing LocalDate::plusDays() method
Hello @BenMorel,

As you've noticed, I'm focusing on some performance aspects of this library.

There's one operation we do very often in our codebase : `LocalDate::plusDays(1)`. Mostly because we often need to convert "date and time" intervals (with an exclusive end) to "date only" intervals (with an inclusive end), and this is done by calling `plusDays(1)` on the end of the interval.

I have a Blackfire inspection that shows that for ~400'000 calls to `plusDays()` in a request, it takes 11.5 seconds, of which :

* 0.5 seconds is spent in the method's code itself
* 4.27 seconds is spent in `LocalDate::toEpochDay`, of which 2.68 seconds in the method's code itself and 1.25 seconds in `intdiv`.
* 6.69 seconds is spent in `LocalDate::ofEpochDay`, of which 4.77 seconds in the method's code itself and 1.92 seconds in `intdiv`.

I don't think there's much that can be done regarding `intdiv`, as it's a native PHP function. And obviously, we can't use a cache to improve performances, as the `LocalDate` we call `plusDays(1)` from will widely vary (aka represent different dates).

So, I'm left wondering if there's maybe something there that could be done within `toEpochDay` and `ofEpochDay` themselves ? But looking at the code, I'm quite lost regarding what could actually be improved, except trying to reduce the number of mathematical operations that are done while keeping all the edge-cases they have to deal with...

Or could it be simplified by having a separate method that does specifically `plusOneDay` (and not "N" days) ?

I'm eager to read your thoughts and insights about this. Thanks !
gnutix

gnutix:
Fixes #78.

Local PHP script-based benchmark shows that for 10'000'000 iterations of `LocalDate::plusDays(1)` (from 2023-09-15 to 29402-10-11) takes an average of 25 seconds without this patch, and an average of 5 seconds with it. Adding the patch does not change the results if the input argument of `plusDays()` is something else than 1 (or only a non-noticeable amount).

<details>
  <summary>Benchmark script</summary>

  ```php
<?php # var/script.php

declare(strict_types=1);

namespace Script;

use Brick\DateTime\LocalDate;
use Brick\DateTime\Stopwatch;
use Symfony\Component\ErrorHandler\Debug;

require dirname(__DIR__) . '/config/bootstrap.php';

Debug::enable();

$stopwatch = new Stopwatch();

$date = LocalDate::parse('2023-09-15');

$stopwatch->start();
for ($i = 0; $i < 10_000_000; ++$i) {
    $date = $date->plusDays(2);

    if ($i === 0 || $i === 9_999_999) {
        dump((string) $date);
    }
}
$stopwatch->stop();

dd((string) $stopwatch->getElapsedTime());
```
</details>

I understand this is quite a specific patch, mostly related to our own use case in our code base, but I still think it might benefit users of this library, because looping from day to day seems like quite a common use case (it's done even within the library, in `LocalDateRange::getIterator()` and `LocalDate::plusWeekdays()`).

We thought of adding a `-1` use case too, but it's way more complicated to implement, and I'd guess it's not as useful/impactful.

Credit goes to [@BastienClement](https://github.com/BastienClement).

### CODE REVIEW

File: src/LocalDate.php [hunk 1]
```diff
@@ -456,6 +456,12 @@ public function plusDays(int $days): LocalDate
             return $this;
         }
 
+        if ($days === 1) {
+            return $this->day >= 28 && $this->day === $this->getLengthOfMonth()
```
BenMorel, line 460: `$this->day >= 28` is redundant with the second condition, I guess this is another optimization to avoid calling `getLengthOfMonth()` needlessly?

src/LocalDate.php
```diff
@@ -456,6 +456,12 @@ public function plusDays(int $days): LocalDate
             return $this;
         }
 
+        if ($days === 1) {
```
BenMorel, line 459: Could you please add a comment to explain why this code is there?

```suggestion
        // Performance optimization for a common use case.
        if ($days === 1) {
```

BenMorel:
Looks good! I agree this is a very common operation, and deserves the x5 performance boost this extra code brings.

src/LocalDate.php
gnutix: It is indeed. 28 times out of 31, we can prevent a method call with a simple math condition. Seemed like a good deal. ;)

gnutix:
I don't think there any more performance improvements I'd like to do for now. Would you mind creating a new minor release once this PR is merged ? :)

gnutix:
(Edit: there's one more performance improvement I'd like to explore before making a release)

BenMorel:
@gnutix Released as [0.5.2](https://github.com/brick/date-time/releases/tag/0.5.2)! Thanks for your contribution.

gnutix:
Thanks! Would you mind adding @BastienClement in the release note as a contributor (if you can edit it) ? He did at least as much as I did for these 3 PRs. :)

BenMorel:
@gnutix No problem, done!

https://github.com/brick/date-time/releases/tag/0.5.2

PR #82, opened 2023-09-24
Call PHPUnit methods through `self::` instead of `$this->`

tigitz:
This is for preference.

Related to https://github.com/brick/date-time/pull/77#issuecomment-1721904161

Note that custom `assertX` functions are not supported by the fixer, it only take into account PHPUnit ones.

I had to manually replace them to maintain consistency.


BenMorel (approved):

BenMorel:
Thank you, @tigitz!

PR #83, opened 2023-09-27
LocalDateRange::toInterval

solodkiy:
Converting `LocalDateRange` to `Interval` is a common task in my app. Typically, I have a method to generate reports that expects a `LocalDateRange` argument. Subsequently, I need to transform it for SQL queries using the `BETWEEN` operator. It’s important to note that the timezone in my database differs from the implied timezone in `LocalDateRange`.

The new method introduced will transform any `LocalDateRange` into an `Interval` in the specified timezone.

### DIFF
--- a/src/LocalDateRange.php
+++ b/src/LocalDateRange.php
@@ -245,6 +245,25 @@ final class LocalDateRange implements IteratorAggregate, Countable, JsonSerializ
     }
 
     /**
+     * Converts this LocalDateRange to Interval instance.
+     *
+     * The result is Interval from 00:00 start date and 00:00 end date + one day (because end in Interval is exclude)
+     * in the given time-zone.
+     */
+    public function toInterval(TimeZone $timeZone): Interval
+    {
+        $startZonedDateTime = $this->getStart()
+            ->atTime(LocalTime::min())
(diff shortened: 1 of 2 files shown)


src/LocalDateRange.php
```diff
@@ -244,6 +244,25 @@ public function jsonSerialize(): string
         return (string) $this;
     }
 
+    /**
+     * Converts this LocalDateRange to Interval instance.
+     *
+     * The result is Interval from 00:00 start date and 00:00 end date + one day (because end in Interval is exclude)
+     * in the given time-zone.
+     */
```
BenMorel, line 252:
```suggestion
    /**
     * Converts this LocalDateRange to an Interval in the given TimeZone.
     *
     * The result is an Interval from 00:00 on the start date to 00:00 on the end date + one day
     * (the end Instant of Interval being exclusive).
     */
```

BenMorel: Also, should we add a warning relative to [this](https://github.com/brick/date-time/blob/0.5.3/src/ZonedDateTime.php#L60-L86)?

As to what happens if `00:00` falls within a DST transition (I'm not even sure if this is possible in any timezone).

Issue #73, opened 2023-07-28
`null`-friendly methods


gnutix:
In our application, we often times have to create Brick objects (`LocalDate`, `Duration`, `TimeZone`, etc) with code like this : 

```php
null !== $var ? Duration::parse($var) : null
```

It would be nice if methods like parse-rs would allow `null` in input, and directly return it if it is null indeed. It would go along with a PHPDoc stating `@return ($input is null ? null : self)`, so that static analysers can tell which result it will be. And that way, no more bloated ternaries at call-site.

WDYT ? I'd be willing to work on a pull request in the upcoming two weeks, if that's something you would accept in the library. Let me know!

gnutix:
In the same vein, we have this `DurationHelper` class that allows us to have `plus` and `minus` null-friendly operations : 

```php
final class DurationHelper
{
    /**
     * @return ($a is null ? ($b is null ? null : Duration) : Duration)
     */
    public static function minus(?Duration $a, ?Duration $b): ?Duration
    {
        /** {@see Duration::minus()} does exactly this, so there's no point trying to optimize functions calls more */
        return self::plus($a, $b?->negated());
    }

    /**
     * @return ($a is null ? ($b is null ? null : Duration) : Duration)
     */
    public static function plus(?Duration $a, ?Duration $b): ?Duration
    {
        if (null === $a && null === $b) {
            return null;
        }
        if (null === $a) {
            return $b;
        }
        if (null === $b) {
            return $a;
        }

        return $a->plus($b);
    }
}
```

It would be much nicer if that would be integrated into the library directly.

gnutix:
Still in the same vein, we use `$date?->jsonSerialize()` a lot when a date might be `null`, but if it's not, we want its ISO-8601 representation. Which is quite ugly, because it's not at all related to JSON serialization. But yet I find it less ugly than `$date?->__toString()` (:vomiting_face:).

So I'm wondering if we could we add a `public function iso(): string` method that would return the same value as `__toString()` ? `$date?->iso()` looks nice.

BenMorel:
Hi, while I understand the use case, I don't think this is something that belongs to this library. A thin helper class is probably fine for this!

Regarding `__toString()`, it may look a bit ugly to call with the null-safe operator, but I'm a bit reluctant to add yet another method that does the same thing :/

gnutix:
[...]

PR #85, opened 2023-09-29
Improve performances for __toString() methods.

gnutix:
Hello @BenMorel,

Another performance PR. `__toString()` can be called a lot in our code, especially on `LocalDate` and `LocalTime` (through `LocalDateTime` and our own `LocalDateInterval`/`LocalDateTimeInterval` classes). Like, this much : 

image

So I've spend a few hours doing lots of experimentation and profiling to find the fastest way to produce the ISO-8601 representation of these objects. The highlights of which are : 

1. concatenation is almost 2 times faster than sprintf :rocket:
2. when padding is needed, str_pad is about 5% faster than sprintf
3. optimizing for the most common path (putting more expensive code behind edge-case conditions helps, for example `$this->year < 1000`)
4. conditions like `$this->month < 10 ? '0' . $this->month : $this->month` are about 30% faster than `substr('0' . $this->month, -2)` (another idea we've tried with @BastienClement to replace `sprintf` calls)

I also checked other classes' `__toString()` methods, and optimized those of `MonthDay`, `YearMonth` and `YearWeek` (which have the same needs as `LocalDate`).

~Finally, I've fixed some bugs in `YearMonth::__toString()`, which was doing `sprintf('%02u', $this->year)` (which seems wrong).~ fixed in #87

And added some more tests where I felt it was needed.

If you merge it, it would be great if you could make a minor release directly. :pray:


src/LocalTime.php
```diff
@@ -662,16 +663,12 @@ public function jsonSerialize(): string
...
-        if ($this->nano === 0) {
-            if ($this->second === 0) {
-                return sprintf('%02u:%02u', $this->hour, $this->minute);
-            } else {
-                return sprintf('%02u:%02u:%02u', $this->hour, $this->minute, $this->second);
-            }
-        }
-
-        $nanos = rtrim(sprintf('%09u', $this->nano), '0');
+        $hasNanos = $this->nano !== 0;
```
gnutix, line 666: Here we could go with `> 0` too. I'm not sure what I prefer. @BenMorel An opinion ?

BenMorel:
Hi @gnutix, thanks a lot for the time you've spent trying new optimizations!

Your tricks are clever but they hurt readability quite a lot :confused:

Before we move forward, one question: are you casting to string many *different* objects? Could caching `__toString()`'s output provide more benefit that these micro-optimizations?

gnutix:
> Your tricks are clever but they hurt readability quite a lot 😕

They are definitely not how anyone would write it at first, that true. I thought about adding a comment "This code is optimized for performances" above it, to keep a trace in the future.

> are you casting to string many different objects? Could caching __toString()'s output provide more benefit that these micro-optimizations?

I've given it a lot of thoughts already, and surely it would be a great benefit (though I don't think both optimizations are exclusive). However, the implementation is a PITA. The obvious way would be to have a private `iso` property that would be set once when `__toString` is first called and then act as a cache. The obscure downfall is that it makes comparing objects fragile : 

```php
$date1 = LocalDate::of(2023, 9, 29);
$date2 = LocalDate::of(2023, 9, 29);
assertEquals($date1, $date2); // true
echo (string) $date2;
assertEquals($date1, $date2); // false : with a diff on the `iso` private property.
```

I've discovered this because I wanted to do exactly that for `LocalDateTimeInterval::isEmpty()` (which is called a lot in our project too).

A way to circumvent this would be to *always* set the `iso` property, like in the constructor. But then the cost of creating that ISO representation is always there, even when not needed, and I'm not sure that's any better.

So.. if you have an idea how we could implement this, I'd be very happy to explore it. Meanwhile, I've opened a PR for something I knew how to fix. :)

BenMorel:
> I thought about adding a comment "This code is optimized for performances" above it. 

I would do that, too.

> At the scale we're at, we're creating/casting both new objects and the same objects multiple times a fair share, so surely it would be a great benefit too (I don't think both optimizations are exclusive though 😉 ).

Definitely not mutually exclusive, but I was wondering which one would affect you the most!

> The obvious way would be to have a private `iso` property that would be set once when `__toString` is first called and then act as a cache.

That's how I would have implemented it, too.

> The obscure downside is that it makes comparing objects fragile :

This is why I always advocate to *not* use loose comparison (`==` or `assertEquals`) on objects, as you're relying on implementation details instead of the public API. `assertTrue($a->isEqualTo($b), "$a != $b")` is the way to go.

> A way to circumvent this downside would be to always set the iso property, in the constructor. But then the cost of creating that ISO representation is always there, even when not needed, and I'm not sure that's an acceptable solution either.

I don't think that's an acceptable solution either.

> A second approach would be to use a WeakMap (so it's garbage collectable), store the object as key, and the ISO representation as value.

That's an interesting solution, where would you put the `WeakMap`? In an external singleton service? As a static property? I'd be interested to see this solution benchmarked.

Both the `iso` property and the static property would make [marking the classes as `@psalm-immutable`](https://github.com/brick/money/pull/75) funny I think, though.

> I hope the readability decrease is still acceptable, considering it's an implementation detail hidden in the methods. If not, I can always try to rewrite it in a style closer to `Duration::__toString()`, using variables, conditions and `$string .= '...';` type of concatenations.

I guess it's acceptable indeed. Let me get back to it over the weekend.

gnutix:
> This is why I always advocate to not use loose comparison (== or assertEquals) on objects, as you're relying on implementation details instead of the public API. assertTrue($a->isEqualTo($b), "$a != $b") is the way to go.

Agree, though that can get hard to avoid when you have more "integration tests" going on, and you might compare arrays of `Timeline` objects, containing arrays of `LocalDateTimeInterval`, made of `LocalDateTime`, themselves with `LocalDate`, etc... :see_no_evil: Comparing each property one by one can become very tedious and complicated.

> That's an interesting solution, where would you put the WeakMap? In an external singleton service? As a static property? I'd be interested to see this solution benchmarked.

I'll see what I can do over the weekend too. :)

gnutix:
Oh, I just noticed `WeakMap` is only available PHP >= 8.0, and the library still supports 7.4.

gnutix:
I've tried adding a static array as cache, but there's too not enough differences between concatenating the properties to make the array key than to make the ISO string.

```diff
diff --git a/src/LocalDate.php b/src/LocalDate.php
index 49da41d..1e66486 100644
--- a/src/LocalDate.php
+++ b/src/LocalDate.php
@@ -63,6 +63,11 @@ final class LocalDate implements JsonSerializable
      */
     private int $day;
 
+    /**
+     * @var array<string, string>
+     */
+    private static $isoCache = [];
+
     /**
      * Private constructor. Use of() to obtain an instance.
      *
@@ -762,7 +767,8 @@ final class LocalDate implements JsonSerializable
      */
     public function __toString(): string
     {
-        return ($this->year < 1000 ? ($this->year < 0 ? '-' : '') . str_pad((string) abs($this->year), 4, '0', STR_PAD_LEFT) : $this->year)
+        return self::$isoCache[$this->year . '-' . $this->month . '-' . $this->day] ??=
+            ($this->year < 1000 ? ($this->year < 0 ? '-' : '') . str_pad((string) abs($this->year), 4, '0', STR_PAD_LEFT) : $this->year)
             . '-'
             . ($this->month < 10 ? '0' . $this->month : $this->month)
             . '-'
```

5 million calls to `__toString()` takes (average of 5 runs) : 
* ~3.64s with the cache
* ~3.66s without the cache

gnutix:
Conclusion : I don't know how to add a cache to `__toString()` methods that would really bring good performance improvements and not have risks / issues / drawbacks. :roll_eyes:

BenMorel:
> If I add these private properties on LocalDate, LocalTime and LocalDateTime in our vendor's copy of brick/date-time, here's what happens to our project's test suite :

Wow, you're indeed relying a lot on this. Funnily enough, I once advised a coworker against using `assertEquals()` with two `LocalDate` objects, due to the risk of me introducing a `__toString()` cache at some point. You've been lucky so far!

> Conclusion : an hour and a half later, I still don't know how to add a cache to __toString() that would really bring good performance improvements and not have risks / issues / drawbacks. 😅

OK, let's forget about this for now. I'm still not convinced about whether or not a local cache of computed values is acceptable (I would not consider breaking loose equality a breaking change, but maybe I can be convinced otherwise), but the lack of evidence of a real-world performance improvement (in addition to the extra memory consumed) makes it a no-go for now. your sentence is suspended :slightly_smiling_face:

gnutix:
Our test suite is not directly relying on assertEquals(LocalDate $a, LocalDate $b), but indirectly by comparing objects that have private properties that are themselves LocalDate (or other similar) Value objects.

Glad the bullet is dodged for now. ;)

BenMorel:
> Our test suite is not directly relying on assertEquals(LocalDate $a, LocalDate $b), but indirectly by comparing objects that have private properties that are themselves LocalDate (or other similar) Value objects.

Shouldn't these objects have an `isEqualTo()` method as well?

gnutix:
They probably should. Never took the time to implement a complex method we did not needed, I guess. :man_shrugging:

gnutix:
I pushed 2 commits regarding #73 (as I already fixed a bug in `YearMonth::__toString()` here) adding `toISO()` methods.

* I've skipped `TimeZoneRegion` and `TimeZoneOffset`, as they already have a `getId(): string` method.
* For `Month`, I named the method `getName()`, as `toISO()` felt wrong.

BenMorel:
> I pushed 2 commits adding toISO() methods (see https://github.com/brick/date-time/issues/73) here, as there's a bug fixed for YearMonth::__toString() that I needed.

Could you please move this to a separate PR? This change is self-contained and bloats the current PR. We'll try to get it merged quickly so that you can rebase the current PR.

Also, I think I'm not sold on `toISO()`. I understand that adding `String` to a method that returns a string feels redundant, but:

- it reminds of `__toString()` from PHP and `toString()` from Java, and really is the canonical string representation for the object, so I quite like that we're including this in the method name
- calling it `toISOString()` makes it searchable in the IDE if I start typing `->tostr`; people not knowing that the string representation is called "ISO" will still be able to find the method

For these reasons, I think I would call it `toISOString()` (or maybe, `toIsoString()`) as suggested by @tigitz.

But we can continue this discussion in the other PR!

gnutix:
> Could you please move this to a separate PR?

Done, cf. #87.

gnutix:
@BenMorel Rebased :) Fixed a wrong variable name in `YearMonthTest::testJsonSerialize()` while at it.

tigitz:
@BenMorel LGTM 👍

I tend to preserve acronym case in naming, helps readability IMO.

gnutix:
Hey @BenMorel ! Any chance we could get this merged and released someday soon ? :pray:

src/LocalDate.php
```diff
@@ -759,9 +762,12 @@ public function jsonSerialize(): string
...
     {
-        $pattern = ($this->year < 0 ? '%05d' : '%04d') . '-%02d-%02d';
-
-        return sprintf($pattern, $this->year, $this->month, $this->day);
+        // This code is optimized for high performance
+        return ($this->year < 1000 ? ($this->year < 0 ? '-' : '') . str_pad((string) abs($this->year), 4, '0', STR_PAD_LEFT) : $this->year)
+            . '-'
+            . ($this->month < 10 ? '0' . $this->month : $this->month)
+            . '-'
+            . ($this->day < 10 ? '0' . $this->day : $this->day);
```
BenMorel, line 770: I think this can be slightly optimized further:

```suggestion
        return (
            ($this->year < 1000 && $this->year > -1000)
                ? ($this->year < 0
                    ? '-' . str_pad((string) -$this->year, 4, '0', STR_PAD_LEFT)
                    : str_pad((string) $this->year, 4, '0', STR_PAD_LEFT)
                )
                : $this->year
            )
            . '-'
            . ($this->month < 10 ? '0' . $this->month : $this->month)
            . '-'
            . ($this->day < 10 ? '0' . $this->day : $this->day);
```

src/LocalTime.php
BenMorel: I would go with `!== 0`, and I would inline the condition. I suspect this may be (very slightly) faster, as well.

src/LocalTime.php
```diff
@@ -662,17 +663,14 @@ public function jsonSerialize(): string
...
 
-        return sprintf('%02u:%02u:%02u.%s', $this->hour, $this->minute, $this->second, $nanos);
+        // This code is optimized for high performance
+        return ($this->hour < 10 ? '0' . $this->hour : $this->hour)
+            . ':'
+            . ($this->minute < 10 ? '0' . $this->minute : $this->minute)
+            . ($hasNanos || $this->second !== 0 ? ':' . str_pad((string) $this->second, 2, '0', STR_PAD_LEFT) : '')
```
BenMorel, line 672: Also, I would put `second` first for better readability:

```suggestion
            . ($this->second !== 0 || $this->nano !== 0 ? ':' . str_pad((string) $this->second, 2, '0', STR_PAD_LEFT) : '')
```

It reads better as hour, minute, second, nano, in this order IMO.

BenMorel: Oh also, why don't you use the same trick as for `hour` and `minute`, to avoid an `str_pad()` call?

```suggestion
            . ($this->second !== 0 || $this->nano !== 0 ? ':' . ($this->second < 10 ? '0' . $this->second : $this->second) : '')
```

BenMorel: Finally, could you please add a test to `LocalTimeTest` to ensure that hours & minutes are always displayed even where they're `0`? I think these are the only tests missing to cover all edge cases.

src/YearWeek.php
```diff
@@ -308,9 +311,10 @@ public function jsonSerialize(): string
      */
     public function toISOString(): string
     {
-        $pattern = ($this->year < 0 ? '%05d' : '%04d') . '-W%02d';
-
-        return sprintf($pattern, $this->year, $this->week);
+        // This code is optimized for high performance
+        return ($this->year < 1000 ? ($this->year < 0 ? '-' : '') . str_pad((string) abs($this->year), 4, '0', STR_PAD_LEFT) : $this->year)
```
BenMorel, line 315: Same logic as for `LocalDate` for the year as well!

src/YearMonth.php
```diff
@@ -300,9 +303,10 @@ public function jsonSerialize(): string
      */
     public function toISOString(): string
     {
-        $pattern = ($this->year < 0 ? '%05d' : '%04d') . '-%02d';
-
-        return sprintf($pattern, $this->year, $this->month);
+        // This code is optimized for high performance
+        return ($this->year < 1000 ? ($this->year < 0 ? '-' : '') . str_pad((string) abs($this->year), 4, '0', STR_PAD_LEFT) : $this->year)
```
BenMorel, line 307: I'd suggest the same logic as for `LocalDate` for the year!

BenMorel (changes requested):
@gnutix Sorry for the delay, after these changes I think we'll be ready to merge!

src/LocalDate.php
gnutix: I don't think I agree with this change ; on the upside, it removes the call to `abs()` (which I've not noticed to be a performance bottleneck) inside an edge-case branch, but on the downside it adds a second condition `&& $this->year > -1000` that will be executed for every call to the method (and it's more verbose/hard to follow IMHO, with yet one more nested ternary).

As 99.99% of dates used in today's web applications will be between 1950 and now+50 years, that change goes against the idea of "optimizing for the most common path and putting more expensive code behind edge-case conditions".

gnutix:
@BenMorel I've pushed all the requested changes, except the code regarding the year's calculation, for which I'm not convinced.

src/LocalDate.php
BenMorel: While I sympathize with the readability argument, I think we've already killed readability, so why not push it to the limits?

> but on the downside it adds a second condition && $this->year > -1000 that will be executed for every call to the method.

No, the second condition [won't be executed](https://3v4l.org/E4T1M) if the first condition is `false`, so it won't add any overhead for the common case.

@tigitz Do you have an opinion on this?

gnutix: Oh right, my bad! It's an "and" and not an "or"... Okay, I'll apply this change then.

For the readability argument, I liked that it was still fitting one line per part (year, month, etc). With this one I'll try to find how to format it best.

gnutix:
@BenMorel There you go, I've applied the last change for the year's code.

BenMorel:
Thank you, @gnutix! [0.5.4](https://github.com/brick/date-time/releases/tag/0.5.4) just released.

PR #86, opened 2023-09-30
Add parsing methods where missing.

### LINKED ISSUE #68: Some objects with ISO representations are missing a parse method
Hello there,

I've noticed that `YearWeek` and other objects have a `__toString()` method returning an ISO representation of the object, but not the corresponding `public static function parse(string $iso): self` method. Which leads to multiple usages of `*::of()` methods sprouting around in our code (some using `substr`, some `explode('-W')`, etc... unpredictable code, no proper error management... nasty things).

Is that something you would be willing to accept as a PR?

Thanks for your answer.
gnutix

gnutix:
Fixes #68 by adding : 

* `YearWeek::parse('2023-W39');`
* `Year::parse('2023');`
*  ~`DayOfWeek::parse('06');`~
* ~`Month::parse('October');`~

I was not sure it made sense to add `Year::parse()`, because it's so close to `Year::of((int) $string)`. But then I realized they do differ for dates (positive and negative) under 1000 / 100 / 10 because of the trailing zeroes : 

```php
Year::of((int) '-100'); // works
Year::of((int) '-0100'); // works, interpreted as -100 (PHP, right? :>)
Year::parse('-100'); // should throw an Exception, as the valid ISO is "-0100"
Year::parse('-0100'); // works
```

~Same for `DayOfWeek` with its leading zero. I know `DayOfMonth` and `Month` will be converted into Enums in the next version ([0.6](https://github.com/brick/date-time/tree/v0.6?rgh-link-date=2023-09-14T16%3A04%3A10Z)), but I still think it doesn't hurt to add these methods on them in the meantime.~

### DIFF
--- a/tests/YearTest.php
+++ b/tests/YearTest.php
@@ -35,6 +35,46 @@ class YearTest extends AbstractTestCase
         Year::of($invalidYear);
     }
 
+    /**
+     * @dataProvider providerParse
+     */
+    public function testParse(string $string, int $expectedYear): void
+    {
+        self::assertYearIs($expectedYear, Year::parse($string));
+    }
+
+    public function providerParse(): array
+    {
+        return [
+            ['-2023', -2023],
+            ['-0100', -100],
+            ['1987', 1987],
+            ['121241', 121241],
+        ];
+    }
+
+    /**
+     * @dataProvider providerParseInvalidYearThrowsException
+     */
+    public function testParseInvalidYearThrowsException(string $invalidValue): void
+    {
+        $this->expectException(DateTimeException::class);
+        $this->expectExceptionMessage('Failed to parse "' . $invalidValue . '"');
+
+        Year::parse($invalidValue);
+    }
+
+    public function providerParseInvalidYearThrowsException(): array
+    {
+        return [
+            [''],
+            ['+2000'],
+            ['-100'],
+            ['ABC'],
+            ['9999999999'],
+        ];
+    }
+
     public function providerOfInvalidYearThrowsException(): array
     {
         return [
--- a/tests/YearWeekTest.php
+++ b/tests/YearWeekTest.php
@@ -336,6 +336,49 @@ class YearWeekTest extends AbstractTestCase
         self::assertSame($expected, (string) $yearWeek);
     }
 
+    /**
+     * @dataProvider providerParse
+     */
+    public function testParse(string $string, int $expectedYear, int $expectedWeek): void
+    {
+        $yearWeek = YearWeek::parse($string);
+        self::assertYearWeekIs($expectedYear, $expectedWeek, $yearWeek);
+    }
+
+    public function providerParse(): array
+    {
+        return [
+            ['-2000-W12', -2000, 12],
+            ['-0100-W01', -100, 1],
+            ['2015-W01', 2015, 1],
+            ['2015-W48', 2015, 48],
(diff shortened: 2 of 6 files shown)


gnutix:
`Instant` and `Interval` don't have parse methods either. Not sure if it would make sense to add them or not. WDYT ?

gnutix:
I did `Month` with a simpler implementation, no Parser and `::from(DateTimeParseResult $result)` method. Let me know if that works for you, or if I should do it differently.

tests/YearTest.php
```diff
@@ -35,6 +35,24 @@ public function testOfInvalidYearThrowsException(int $invalidYear): void
...
+    public function providerParse(): array
+    {
+        return [
+            ['-2023', -2023],
+            ['-0100', -100],
+            ['1987', 1987],
+            ['121241', 121241],
+        ];
+    }
+
```
BenMorel, line 55: Can you please add a `testParseInvalidYearThrowsException()` with a couple data such as `''`, `'9999999999'`, `'-100'`, `'+2000'`, `'ABC'`?

tests/YearWeekTest.php
```diff
@@ -336,6 +336,26 @@ public function testToString(int $year, int $week, string $expected): void
...
+    {
+        return [
+            ['-2000-W12', -2000, 12],
+            ['-0100-W01', -100, 1],
+            ['2015-W01', 2015, 1],
+            ['2015-W48', 2015, 48],
+            ['120195-W23', 120195, 23],
+        ];
+    }
+
```
BenMorel, line 358: Please add a `testParseInvalidYearWeekThrowsException()` with e.g. `''`, `'+2000-W01'`, `'2000W01`, `'2000-W54'`, `XXXX-W53` with `XXXX` being a 52-week year!

BenMorel:
> `Instant` and `Interval(Instant, Instant)` don't have parse methods either. Not sure if it would make sense to add them or not. WDYT ?

If they're not needed at the moment, I would leave them for now.

BenMorel (changes requested):
@gnutix I'm OK with `Year` and `YearWeek`, but not `Month` and `DayOfWeek` for the reasons stated above!

tests/YearTest.php
gnutix: Done

tests/YearWeekTest.php
gnutix: Done

src/DayOfWeek.php
```diff
@@ -52,6 +56,35 @@ public static function of(int $dayOfWeek): DayOfWeek
...
+     */
+    public static function parse(string $text, ?DateTimeParser $parser = null): DayOfWeek
+    {
+        if (! $parser) {
+            $parser = IsoParsers::dayOfWeek();
+        }
+
+        return DayOfWeek::from($parser->parse($text));
+    }
+
```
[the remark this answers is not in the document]
gnutix: Reverted!

BenMorel:
Thank you, @gnutix!

To iterate faster and avoid yet another feedback, I took the liberty to update your commits to rename data providers that didn't match the project's standards:

- ~providerParseInvalidValues~ `providerParseInvalidYearThrowsException`
- ~providerParseInvalidValues~ `providerParseInvalidYearWeekThrowsException`

Issue #89, opened 2023-10-18
Year::toISOString() not conformation to ISO 8601


andreaswolf:
According to Wikipedia and https://github.com/moment/moment/issues/3842#issuecomment-287738049 (I don't have access to the norm myself, so I need to quote other sources), the year part of any date in ISO 8601 should have at least four digits, with leading zeros for years < 1000. The `PATTERN` constant in `YearField` also reflects this.

The new method `toISOString()`, however, just does a (string) on the integer member `$year`. IMO, this should be handled similar to `LocalDate::toISOString()`.

BenMorel:
I think you're right indeed. We need to copy the year logic from `YearMonth` / `LocalDate`.
Do you want to open a PR?

andreaswolf:
> I think you're right indeed. We need to copy the year logic from `YearMonth` / `LocalDate`. Do you want to open a PR?

Just did that. I had a patch applied locally already, it was just missing tests.

PR #90, opened 2023-10-20
Stringify years correctly according to ISO 8601

andreaswolf:
Resolves: #89

### DIFF
--- a/src/Year.php
+++ b/src/Year.php
@@ -10,6 +10,10 @@ use Brick\DateTime\Parser\DateTimeParseResult;
 use Brick\DateTime\Parser\IsoParsers;
 use JsonSerializable;
 
+use function str_pad;
+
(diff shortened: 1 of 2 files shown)


src/Year.php
```diff
@@ -281,7 +285,13 @@ public function jsonSerialize(): string
      */
     public function toISOString(): string
     {
-        return (string) $this->year;
+        return $this->year < 1000 && $this->year > -1000
```
BenMorel, line 288: Could you please add the comment:

```php
// This code is optimized for high performance
```

to justify that it’s not using sprintf?

andreaswolf: Done

BenMorel:
Thank you!

BenMorel:
Released as [0.5.5](https://github.com/brick/date-time/releases/tag/0.5.5).

PR #96, opened 2024-03-22
YearMonth: deprecate int argument for month, add support for Month enum.

gnutix:


src/YearMonth.php
```diff
@@ -156,10 +162,10 @@ public function compareTo(YearMonth $that): int
         if ($this->year > $that->year) {
             return 1;
         }
-        if ($this->month < $that->month) {
+        if ($this->month->value < $that->month->value) {
             return -1;
         }
-        if ($this->month > $that->month) {
+        if ($this->month->value > $that->month->value) {
```
gnutix, line 168: Should there be a `compareTo` method on `Month` ?

BenMorel: I'm not sure if it makes sense to compare months without a year, I'd leave it like this for now!

src/YearMonth.php
```diff
@@ -26,16 +29,12 @@ final class YearMonth implements JsonSerializable, Stringable
      */
     private readonly int $year;
 
-    /**
-     * The month, from 1 to 12.
-     */
-    private readonly int $month;
+    private readonly Month $month;
```
BenMorel, line 32: Should we store the `Month` or its value? We have a precedent with `LocalDate`, that [stores the `int` value and not the enum](https://github.com/brick/date-time/blob/0.6.0/src/LocalDate.php#L59). Not saying this is the right thing to do, but I think we should do the same here for consistency.

Another reason is backwards compatibility: while I'm generally against using it this way, I know that some people compare brick/date-time objects with `==` (or `assertEquals()` in PHPUnit), so this would break here if we're releasing this in a patch release.

We can definitely do this in the next version though (`0.7`), so let's discuss this in another PR/issue if you wish?

gnutix: Sure, didn't thought of the == here. There should be a unit test covering that!

I'll revert it whenever I have some time to do so. And maybe add the test. ;)

BenMorel:
Side note: I was thinking of removing `getMonth()` in the next release (`0.7`), to re-introduce it later in `0.8` as returning a `Month` enum.

The other option would be to change the signature right away in `0.7`, but I feel like skipping a version makes it more likely for people upgrading one version at a time to notice the change. Do you have an opinion on this?

gnutix:
> Side note: I was thinking of removing `getMonth()` in the next release (`0.7`), to re-introduce it later in `0.8` as returning a `Month` enum.
> 
> The other option would be to change the signature right away in `0.7`, but I feel like skipping a version makes it more likely for people upgrading one version at a time to notice the change. Do you have an opinion on this?

I have no experience releasing open source libraries. But if it's a B's break, shouldn't it be done in a new major version, like 1.0 ? I was wondering why you don't have such a version yet, don't you think this library is stable / production-ready ? (I do)

src/YearMonth.php
gnutix: Done!

BenMorel:
> But if it's a BC Break, shouldn't it be done in a new major version, like 1.0 ? I was wondering why you don't have such a version yet, don't you think this library is stable / production-ready ? (I do)

[semver](https://semver.org/) allows breaking changes in version `0.x`. I keep non-breaking changes in patch releases, and breaking changes in minor releases.

The library is definitely production quality, but I feel like the API is not stable yet. Once we reach the `1.0.0` release, we will have to release a major version for even the smallest of breaking changes. If I had released `1.0.0` from the beginning, we would be at version `6.0.0` already, with sometimes very few minor/patch releases in between major versions.

Tagging `1.0.0` one day, will mean that I consider the library mostly finished and the API stable.

> Sure, didn't thought of the == here. There should be a unit test covering that!

Adding a unit test would mean that we support this use case, and I'm personally against using objects this way. Loose comparison is an unfortunate "feature" of PHP, which exposes object internals, and supporting it prevents using features like internal cache / lazy loading. We're a bit in a grey zone here, where I want to avoid the BC break *if possible*, but still not officially support this use case.

src/YearMonth.php
```diff
@@ -44,17 +47,25 @@ private function __construct(int $year, int $month)
...
 
-        return new YearMonth($year, $month);
+        if (is_int($month)) {
+            // usually we don't use trigger_error() for deprecations, but we can't rely on @deprecated for a parameter type change;
+            // maybe we should revisit using trigger_error() unconditionally for deprecations in the future.
+            trigger_error('Passing an integer to YearMonth::of() second argument is deprecated, pass a Month instance instead.', E_USER_DEPRECATED);
+
+            Field\MonthOfYear::check($month);
+
+            $month = Month::from($month);
```
BenMorel, line 65: Shouldn't you return `new YearMonth($year, $month);` straight away here? Otherwise you're instantiating an enum for nothing!

src/YearMonth.php
```diff
@@ -209,17 +220,25 @@ public function withYear(int $year): YearMonth
...
+            // usually we don't use trigger_error() for deprecations, but we can't rely on @deprecated for a parameter type change;
+            // maybe we should revisit using trigger_error() unconditionally for deprecations in the future.
+            trigger_error('Passing an integer to YearMonth::withMonth() is deprecated, pass a Month instance instead.', E_USER_DEPRECATED);
+
+            Field\MonthOfYear::check($month);
+
+            $month = Month::from($month);
+        }
+
+        $month = $month->value;
```
BenMorel, line 236:
```suggestion
            Field\MonthOfYear::check($month);
        } else {
            $month = $month->value;
        }

        $month = $month->value;
```

BenMorel (changes requested):

BenMorel (approved):

BenMorel:
Thank you, @gnutix!

PR #98, opened 2024-03-22
Stopwatch Lap Duration

rodnaph:
Return the lap duration when stopping the `Stopwatch`, which will be the duration since the last start.

If the stopwatch is not started, then the duration will be zero.

Use case - using a stopwatch to perform multiple duration timings, and then fetch the total elapsed time at the end (I can't see this is possible now, please direct me if I've just overlooked it though).

**Note:** CI failure does not seem related to my changes, as above though let me know if I missed something.

### DIFF
--- a/src/Stopwatch.php
+++ b/src/Stopwatch.php
@@ -49,14 +49,14 @@ final class Stopwatch
     }
 
     /**
-     * Stops the timer.
+     * Stops the timer and returns the lap duration.
      *
-     * If the timer is already stopped, this method does nothing.
+     * If the timer is already stopped, this method does nothing, and returns a zero duration.
      */
-    public function stop(): void
+    public function stop(): Duration
     {
         if ($this->startTime === null) {
-            return;
+            return Duration::zero();
         }
 
         $endTime = $this->clock->getTime();
@@ -64,6 +64,8 @@ final class Stopwatch
 
         $this->duration = $this->duration->plus($duration);
         $this->startTime = null;
+
+        return $duration;
     }
 
     /**
--- a/tests/StopwatchTest.php
+++ b/tests/StopwatchTest.php
@@ -73,8 +73,10 @@ class StopwatchTest extends AbstractTestCase
     public function testStopWithNullStartTime(): void
     {
         $stopwatch = new Stopwatch();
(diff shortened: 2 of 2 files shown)


gnutix:
Sorry for the CI failure, it's fixed by #100

BenMorel:
Hi, I like this idea, this feature is indeed missing!

What about adding a `getLapTime()` as well, to be able to get the duration since the last start without stopping?

By the way you can rebase now, this will fix the CS issue.

rodnaph:
> What about adding a `getLapTime()` as well, to be able to get the duration since the last start without stopping?

Thanks, I've added that method, let me know if there's anything I've missed.

BenMorel (changes requested):

src/Stopwatch.php
```diff
@@ -81,6 +81,15 @@ public function isRunning(): bool
         return $this->startTime !== null;
     }
 
+    public function getLapTime(): Duration
+    {
+        if ($this->startTime === null) {
+            return Duration::zero();
```
BenMorel, line 87: Actually I looked for other `Stopwatch` implementations, and stumbled upon the [JBoss](https://docs.jboss.org/jbossas/javadoc/4.0.4/common/index.html?org/jboss/util/StopWatch.html) one in Java, that looks like a good one. This one memorizes the lap duration after the stopwatch was stopped, and returns the previous value and not zero.

Rather that rushing to add this feature, I think I'd rather pass on this one until I have a good idea of what a good implementation would be. Would you mind removing it for now? Sorry for the waste of time :confused:

BenMorel:
Apart from this, returning the duration in `stop()` looks good to me, @rodnaph. Let's fix the nitpicks and merge it!

src/Stopwatch.php
rodnaph: No problem, I've removed that method and tried to address your other comments, let me know if I missed something, thanks.

BenMorel (approved):

BenMorel:
Thank you, @rodnaph!

BenMorel:
Released as [0.6.3](https://github.com/brick/date-time/releases/tag/0.6.3).

PR #99, opened 2024-03-22
Install Rector and apply ClassPropertyAssignToConstructorPromotionRector

gnutix:
Hello there! I've noticed that the project was not using promoted properties, even though its minimum PHP version supports them. Yet it's using readonly properties already. So I thought it would be a good idea to add Rector in the project to allow doing automated refactorings to keep the codebase up-to-date.

With the current config file, if you bump the minimum PHP version in `composer.json`, it will apply the refactorings that can be done in the new PHP version, making sure the code is always the latest and brightest.

cf. https://getrector.com/documentation/integration-to-new-project

Let me know what you think.

### TODO
- [x] Install Rector in `tools/` (to be run with `tools/rector/vendor/bin/rector process -c tools/rector/rector.php`)
- [x] PHP 5.3 - 5.6 (nothing changed)
- [x] PHP 7.0 - 7.4 (nothing changed)
- [x] PHP 8.0 :
  - [x] `ClassPropertyAssignToConstructorPromotionRector` applied (I made my best to keep the most complete comments around)
- [x] PHP 8.1 (nothing changed)
- [x] Add type coverage at maximum level
- [x] Add dead code at maximum level
  - [x] ~`SimplifyUselessVariableRector`~ skipped
- [x] ECS: apply `TrailingCommaInMultilineFixer` for everything

### DIFF
--- a/src/Clock/FixedClock.php
+++ b/src/Clock/FixedClock.php
@@ -16,8 +16,9 @@ final class FixedClock implements Clock
     /**
      * @param Instant $instant The time to set the clock at.
      */
-    public function __construct(private Instant $instant)
-    {
+    public function __construct(
+        private Instant $instant
+    ) {
     }
 
     public function getTime(): Instant
--- a/src/Clock/FixedClock.php
+++ b/src/Clock/FixedClock.php
@@ -17,7 +17,7 @@ final class FixedClock implements Clock
      * @param Instant $instant The time to set the clock at.
      */
     public function __construct(
-        private Instant $instant
+        private Instant $instant,
     ) {
     }
 
--- a/src/Clock/FixedClock.php
+++ b/src/Clock/FixedClock.php
@@ -13,14 +13,11 @@ use Brick\DateTime\Instant;
  */
 final class FixedClock implements Clock
 {
-    private Instant $instant;
-
     /**
      * @param Instant $instant The time to set the clock at.
      */
-    public function __construct(Instant $instant)
+    public function __construct(private Instant $instant)
     {
-        $this->instant = $instant;
     }
 
     public function getTime(): Instant
--- a/src/Clock/ScaleClock.php
+++ b/src/Clock/ScaleClock.php
@@ -14,19 +14,9 @@ use Brick\DateTime\Instant;
 final class ScaleClock implements Clock
 {
     /**
-     * The reference clock.
-     */
-    private readonly Clock $referenceClock;
-
-    /**
      * The start time.
      */
     private readonly Instant $startTime;
-
-    /**
-     * The time scale.
-     */
-    private readonly int $timeScale;
 
     /**
      * - a scale > 1 makes the time move at an accelerated pace;
@@ -37,11 +27,11 @@ final class ScaleClock implements Clock
      * @param Clock $referenceClock The reference clock.
      * @param int   $timeScale      The time scale.
      */
-    public function __construct(Clock $referenceClock, int $timeScale)
-    {
-        $this->referenceClock = $referenceClock;
-        $this->startTime = $referenceClock->getTime();
-        $this->timeScale = $timeScale;
+    public function __construct(
+        private readonly Clock $referenceClock,
+        private readonly int $timeScale
+    ) {
+        $this->startTime = $this->referenceClock->getTime();
     }
 
     public function getTime(): Instant
--- a/src/Clock/ScaleClock.php
+++ b/src/Clock/ScaleClock.php
@@ -29,7 +29,7 @@ final class ScaleClock implements Clock
      */
     public function __construct(
         private readonly Clock $referenceClock,
-        private readonly int $timeScale
+        private readonly int $timeScale,
     ) {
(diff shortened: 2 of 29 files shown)


gnutix:
> Coverage decreased (-0.01%) to 99.363%

Sorry for that :smile:

src/Clock/FixedClock.php
```diff
@@ -13,14 +13,11 @@
  */
 final class FixedClock implements Clock
 {
-    private Instant $instant;
-
     /**
      * @param Instant $instant The time to set the clock at.
      */
-    public function __construct(Instant $instant)
+    public function __construct(private Instant $instant)
```
BenMorel, line 19: Shouldn't `StandaloneLinePromotedPropertyFixer` fix this?

src/Clock/ScaleClock.php
```diff
@@ -13,21 +13,11 @@
...
 {
-    /**
-     * The reference clock.
-     */
-    private readonly Clock $referenceClock;
-
     /**
      * The start time.
      */
     private readonly Instant $startTime;
```
BenMorel, line 19: While I'm quite a heavy user of promoted properties these days, I'm still a bit annoyed when I see a mix of promoted an non-promoted properties, I feel like this makes the class a bit harder to read. WDYT?

src/Duration.php
```diff
@@ -539,9 +539,8 @@ public function getNanos(): int
     public function getTotalMillis(): int
     {
         $millis = $this->seconds * 1000;
-        $millis += intdiv($this->nanos, 1_000_000);
 
-        return $millis;
+        return $millis + intdiv($this->nanos, 1_000_000);
     }
```
BenMorel, line 544: I'm not fond of this one (and any of those in this commit). I'd either return the whole operation without any variable (on several lines for more readability), or leave it this way.

The way this rule if currently applied, I feel like we're doing one half of a job!

tests/AbstractTestCase.php
```diff
@@ -10,7 +10,6 @@
 use Brick\DateTime\LocalDateRange;
 use Brick\DateTime\LocalDateTime;
 use Brick\DateTime\LocalTime;
-use Brick\DateTime\Month;
```
BenMorel, line 13: How is this not caught by ECS? :thinking:

src/Clock/FixedClock.php
gnutix: It only make them on their own lines when there are more than one. I've checked if there's an option for this behavior, but there's none.

Would you like me to change that by hand ?

src/Clock/ScaleClock.php
gnutix: I've gotten used to it, and in a sense, I like that it separates "dependencies" of an object (what it needs in order to be created) and "internal properties" (used for logic, shortcuts, etc).

src/Duration.php
gnutix: I thought this one commit might clash with the current "code style" of the library. I will revert it and skip this rule in `rector.php`.

tests/AbstractTestCase.php
gnutix: Good question... I've had my share of unused imports not being picked up by ECS, but I've never understood how it comes to be. Never bothered to open an issue, as I'm working on a private project for which I can't just share the whole codebase to help reproduce it.

But maybe it would make sense to create a fork of brick/date-time and share it with @tomasvotruba so he could investigate.

gnutix: There it is : https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/7908

src/Clock/FixedClock.php
```diff
@@ -16,8 +16,9 @@ final class FixedClock implements Clock
     /**
      * @param Instant $instant The time to set the clock at.
      */
-    public function __construct(private Instant $instant)
-    {
+    public function __construct(
+        private Instant $instant
```
BenMorel, line 20: Looks like we need to configure [TrailingCommaInMultilineFixer](https://cs.symfony.com/doc/rules/control_structure/trailing_comma_in_multiline.html) to work not just on arrays!

gnutix: I did it but decided not to commit as I wasn't sure you'd wanted the change. I could do it here or in another PR, as you'd like.

BenMorel: Here is fine!

gnutix:
I think we're all good here. Would you mind creating a release once this PR is merged? :pray:

src/Clock/ScaleClock.php
BenMorel: I agree with you when used in services. In this particular case, I'm less convinced. It's pretty much the only one that bothers me in the codebase though, I guess I'll have to learn to live with it!

BenMorel (approved):

BenMorel:
Thank you, @gnutix!

BenMorel:
@gnutix Released in [0.6.1](https://github.com/brick/date-time/releases/tag/0.6.1)!

PR #100, opened 2024-03-22
Fix ECS errors in Year.php

gnutix:
Apparently #95 broke ECS. Sorry for that.


BenMorel:
Yeah I noticed this right after clicking the merge button. I've changed the repo settings so that the CI must pass before merge, this should prevent this from happening in the future! Thank you :+1:

PR #102, opened 2024-03-27
Fix Month enum argument deprecations within the library itself.

gnutix:
Hello @BenMorel ,

I tried upgrading to 0.6.1 in our project, but it [broke our Behat test suite](https://stackoverflow.com/a/31059663) because the library in itself is now triggering tons of deprecations via `trigger_error()`. Anyway, I tried to "finish the job" of using `Month` wherever there is a month concerned within the library itself, while leaving the usages of the `|int` alternative in the tests to make sure it still works.

I also had to use the `Field\MonthOfYear::check($month);` line *before* creating the Enum to keep the currently throwned `DateTimeException`, otherwise it would throw a PHP `ValueError` when given invalid values to `BackendEnum::from()`.

Let me know what you think. I had to rush it quite a bit, so there's probably room for improvements. I'm not a fan of how the code looks right now, because of this compatibility layer, and even more because we don't store the Month instance internally, forcing to recreate it in a lot of places. But I'm also glad we don't change that, as already discussed. :sweat_smile:

### DIFF
--- a/src/Field/WeekOfYear.php
+++ b/src/Field/WeekOfYear.php
@@ -7,6 +7,7 @@ namespace Brick\DateTime\Field;
 use Brick\DateTime\DateTimeException;
 use Brick\DateTime\DayOfWeek;
 use Brick\DateTime\LocalDate;
+use Brick\DateTime\Month;
 
 /**
  * The week-of-year field.
@@ -51,7 +52,7 @@ final class WeekOfYear
      */
     public static function is53WeekYear(int $year): bool
     {
-        $date = LocalDate::of($year, 1, 1);
+        $date = LocalDate::of($year, Month::JANUARY, 1);
         $dayOfWeek = $date->getDayOfWeek();
 
         return $dayOfWeek === DayOfWeek::THURSDAY
--- a/src/LocalDate.php
+++ b/src/LocalDate.php
@@ -16,9 +16,12 @@ use JsonSerializable;
 use Stringable;
 
 use function intdiv;
+use function is_int;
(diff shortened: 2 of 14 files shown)


BenMorel:
@gnutix, I'm actually thinking of un-deprecating passing an `int` to these functions. After double-checking the original Java implementation, I noticed that they use method overloading to accept both `int` and `Month` whenever a month is accepted. A couple examples:

- [LocalDate](https://docs.oracle.com/javase/8/docs/api/java/time/LocalDate.html):
    - `static LocalDate of(int year, int month, int dayOfMonth)`
    - `static LocalDate of(int year, Month month, int dayOfMonth)`
- [YearMonth](https://docs.oracle.com/javase/8/docs/api/java/time/YearMonth.html):
    - `static YearMonth of(int year, int month)`
    - `static YearMonth of(int year, Month month)`

I feel like this gives greater flexibility to the library, WDYT? (You're still welcome to pass `Month` enums internally as per your PR though, if it makes the code more readable.)

*Side note, I noticed that they strangely only accept `int` in `withMonth()` methods, I'm not sure why. But that shouldn't prevent us from accepting both.*

Finally, could you please do one update per commit? Like:

> - Accept Month in LocalDate::of() & LocalDate::withMonth()
> - Accept Month in MonthDay::of() & MonthDay::withMonth()
> - ...
> - Convert internal calls to Month enum

This will make reviewing the code much easier. Thanks!

gnutix:
There you go : 

* #103
* #104 
* #105 
* #106

PR #105, opened 2024-03-30
LocalDateTime::of() / LocalDateTime::withMonth() / ZonedDateTime::withMonth() allows Month enum

gnutix:


gnutix:
Cannot be separate from #106

